PR #28, opened 2014-05-05
Adds support for OS X without X11.

Sonicadvance1:
OS X support shouldn't depend on X11 being installed on the host.
Removes the X11 dependency.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -78,7 +78,7 @@ case $host_os in
         ;;
     darwin*)
         build_egl=no
-        build_glx=yes
+        build_glx=no
         build_wgl=no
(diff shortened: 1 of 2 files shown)


blmarket:
Too sad to see there is no support for GLX on OS X anymore. personally I'd like to use dwb(keyboard-only web browser, http://portix.bitbucket.org/dwb/), and dependency chain leads me here(dwb depends on webkitgtk, webkitgtk depends on gtk+3 with x11 support, and gtk+3 with x11 depends on GLX from this project...)

Seems I need to create an alternative homebrew formula which revert this in my favor. or do you have any suggestion?

Issue #34, opened 2014-10-23
Add Support to Build libepoxy with Visual Studio


fanc999:
Hi,

From the next stable version of GTK+ (3.16.0, and the current git master branch for it), we would start to use libepoxy so that it would make it simpler to make the OpenGL calls on various systems.

Since GTK+ supports building with Visual Studio, I would like to maintain that every part of the stack is buildable with Visual Studio as well, and I would like to propose the changes to libepoxy to make that happen, as libepoxy currentlydepends on C99 features and gcc-specific directives.

I will attach a series of patches that are needed for this process here, to hopefully make things not-too-hard to maintain.

With blessings, thank you!

anholt:
I'm a terrible maintainer.  I just went looking through the patches now.  On Linux I'm getting 2 fewer tests passing (GLX_NON_APPLE_TEST_PROGS vs _TESTS I think), and I'd like to see a license on detect-msvc.mak, which is quite a bit of code to not have a license.  If you fixed those and force-pushed to your repo, I'd pull it (well, and I'd update COPYING with appropriate info as well).

fanc999:
Hi,

Since I (later had) opened a pull request with issue #35, probably it might be good that further items about this part is directed there, so this becomes more-or-less a duplicate issue.  Will close this issue as a result.

With blessings, thank you!

PR #35, opened 2014-10-29
Enable Build With Visual Studio

fanc999:
Hi,

I thought it might be easier for you to review the patches that are done in the github workflow, so here they are.

Please see issue #34 for details about these.

With blessings, thank you!


anholt:
Alternatively, I'd *really* like to see an MSVC 2013-only port.  I'm really uncomfortable with putting all of this stuff in (particularly the headers with the many different licenses) just to deal with a stale compiler.

fanc999:
Hello Eric,

I see, but then the main reason for this is because we would like to have support for at least Visual Studio 2010 (and Visual Studio 2008, which I believe is more or less the same situation with Visual Studio 2010, except that it dose not ship with stdint.h), which many people that are using and building GTK+ are still using, which is the main driving force behind this.

Perhaps it would be better if I have people load their own stdint.h and/or inttypes.h and/or sdtbool.h, which I think is part of your concerns, instead of bundling them with the sources?  I will do the updates for this part in a bit, and let people know that they need to do so.

Unfortunately this is unlike the xz-utils case where the public headers are still usable from older compilers, even when the sources are C99, as these headers (in particularly the generated ones), have to be updated as well to be usable in older compilers, as the scripts generate both the sources and headers.

With blessings, thank you!

fanc999:
Hello Eric,

I just saw your reply in issue #34 that two fewer tests were passing.  Are they in the tests/?  Somehow I didn't realize them, as I tried "make distcheck" on my Fedora system, and all was well.  I will update the MSVC NMake Makefiles to make sure that they have license comments later on today though.

With blessings, thank you!

fanc999:
Hello Eric,

I removed the headers that are used for C99 compatibility, and instead have the people that will build and use libepoxy acquire them instead, as they are not that hard to find.

About making the rest compatible with only MSVC 2013+, it does seem to me only the Python script can be changed from this point from my patchset, if we really want to go that route.  All the other patches are all still needed for MSVC 2013 (obviously sans the addition of C99-compatibility headers, which I dropped), and even so, it does seem that the update required for the Python script is also largely needed, via some simple tests here using MSVC 2013 with all the latest updates.

The differences in the changes in the Python scripts for pre-MSVC 2013 builds and for MSVC 2013+ builds are actually quite small, compared to the other required changes, and I think it would still be best to keep the public headers usable by older, non-GCC compilers as well, like what xz-utils does, at least on Windows.  Even in MSVC 2013 and later, I don't think having empty structs is a notion that would ever be accepted there.

With blessings, thank you!

nacho:
@fanc999 I guess if we want to go further on this would be better to split this in 2 issues. First get a patch that adds support for msvc 13 and then think about the backward compatibility things. Any chance you could do this?

fanc999:
Hi @nacho,

I will look into splitting this up, as you mentioned.  However, I won't be able to do so until next week as I would be away on a family trip for a vacation.  Thanks for the tip-I think it would make it more apparent that the code changes required for MSVC 2013+ is only just a teeny-meeny little bit less for earlier Visual Studio versions, due to syntactical requirements and the use of GCC-specific items in the current code.

With blessings, thank you!

fanc999:
Hi,

I think I am finally ready to split this pull request up, so I got rid of my fork and re-started this whole thing.

So, I think it is better that I close this pull request and start a new one for MSVC 2013 and another for earlier MSVC versions.

PR #38, opened 2014-12-24
Check if X11 headers are available on MacOS

ebassi:
Instead of unconditionally depending on GLX, we check if the X11 headers
are available, and if not, the GLX build is disabled.


archshift:
Can probably be closed now.

ebassi:
indeed, let's close it.

PR #50, opened 2015-04-28
egl: add EGL_MESA_image_dma_buf_export

elmarco:
This commit adds support for EGL_MESA_image_dma_buf_export from:
http://cgit.freedesktop.org/mesa/mesa/commit/?id=8f7338f284cdb1fef64c85e3293d2200d0cc6387

thanks


airlied:
don't bother with this now, the extension is merged to the registry and I've sent an update for that instead.

PR #55, opened 2015-06-05
Allow Build on Visual Studio 2013

fanc999:
Hi,

As suggested by Eric and Nacho, this is the pull request for allowing libepoxy to build with Visual Studio 2013 and later, and another will be opened (which will depend on this pull request) to support earler Visual Studio versions, as the changes from that point would not be that big.

The following commits on my fork was done so that the code will build with Visual Studio 2013 and later, but does not contain any build files, which would:
-Probably belong to another pull request
-One can come up with such build files rather easily.

With blessings, thank you!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -23,7 +23,7 @@
 AC_PREREQ([2.60])
 AC_INIT([libepoxy],
         [1.2],
-        [https://people.freedesktop.org/~anholt/libepoxy],
+        [https://github.com/anholt/libepoxy],
         [libepoxy])
 AC_CONFIG_SRCDIR([Makefile.am])
 AC_CONFIG_HEADERS([config.h])
--- a/include/epoxy/gl.h
+++ b/include/epoxy/gl.h
@@ -93,9 +93,9 @@ extern "C" {
 
 #include "epoxy/gl_generated.h"
 
-bool epoxy_has_gl_extension(const char *extension);
-bool epoxy_is_desktop_gl(void);
-int epoxy_gl_version(void);
+EPOXY_IMPORTEXPORT bool epoxy_has_gl_extension(const char *extension);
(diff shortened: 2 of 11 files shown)


anholt:
I've cherry-picked most of your changes.  I split up the 4 mashed-together changes for src/ to make sense of things (you'd described them separately in the commit message, anyway).  I dropped one of the changes there, though -- you took a bunch of library-private symbols and marked them as PUBLIC, when they shouldn't be exported at all.  If you need some decoration for internal symbols to the library, the we can look at adding that, but PUBLIC isn't the one.

fanc999:
Hello Eric,

Hmm..., it seems that the problem is with epoxy_conservative_gl_version(), where it is marked as PUBLIC in dispatch_common.c, but the prototype in dispatch_common.h is not marked.  Perhaps also mark the prototype in dispatch_common.h, or maybe just remove the PUBLIC annotation from the function definition in dispatch_common.c, since that function is not used outside of the libepoxy DLL sources?  Please let me know which do you prefer.

With blessings, thank you!

anholt:
Oh, yeah, epoxy_conservative_gl_version() shouldn't be public.

fanc999:
Hi Eric,

Sorry, it took me a bit longer to get back to you...

The last 2 updates that I just had removes the PUBLIC annotation from epoxy_conservative_gl_version() from dispatch_common.c (and dispatch_common.h, did in 2 commits so that it's easier for you to cherry-pick), which will also fix the build on Visual Studio.

I intentionally left the other PUBLIC annotations in dispatch_common.c alone for now, as they might be used elsewhere as well, such as in the tests, though I didn't check for those in particular.

With blessings, thank you!

fanc999:
Hi,

For records, since the changes required for libepoxy to support building under Visual Studio 2013 has been merged, I think it would be good to close this pull request.

With blessings, thank you!

PR #56, opened 2015-06-05
Enable libepoxy to build with Visual Studio 2008-2012

fanc999:
Hi,

As libepoxy is being used with GTK+-3.16.x and later, we would also like to ensure that Visual Studio 2008 and 2012 is supported as well.  It turns out that there are very few changes that are needed to make that happen, once we have made the code work with Visual Studio 2013 (i.e. issue #55).

The first 4 commits are the ones that are needed to make the code work for Visual Studio 2013, which are 100% identical to the ones in issue #55:
-Public Headers: Allow Build On Visual Studio
-src/: Allow Build On Visual Studio
-src/gen_dispatch.py: MSVC 2013: Allow Generated Code To Build
-test/wgl_per_context_funcptrs.c: Fix Build on Visual Studio

Which I was not able to filter out from this issue though, sorry, as this work builds upon the work to make the code buildable with Visual Studio 2013.

With blessings, thank you!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -98,6 +98,7 @@ AC_SUBST(EPOXY_LINK_LIBS)
 
 AM_CONDITIONAL(BUILD_EGL, test x$build_egl = xyes)
 if test x$build_egl = xyes; then
+    PKG_CHECK_MODULES(EGL, [egl])
     AC_DEFINE([BUILD_EGL], [1], [build EGL tests])
 fi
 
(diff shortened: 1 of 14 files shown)


fanc999:
Hi,

For references, since I am not sure how one can filter out the commits that are also done in my other pull request (#55, "Allow Build on Visual Studio 2013"), as this depends on that pull request, the actual commit that makes the difference from that PR would be "src/gen_dispatch.py: Make Generated Code Build Under C89" (commit 440c7d3).

With blessings, thank you!

PR #57, opened 2015-06-16
Fix aclocal problems

xclaesse:


anholt:
There is no m4 directory in the tree, and the patch breaks the build.

xclaesse:
Without the patch, when building on my system it prints this warning (but builds fine):
libtoolize: Consider adding `AC_CONFIG_MACRO_DIR([m4])' to configure.ac and
libtoolize: rerunning libtoolize, to keep the correct libtool macros in-tree.
libtoolize: Consider adding `-I m4' to ACLOCAL_AMFLAGS in Makefile.am.

When building in cerbero (GStreamer build env) environment however it fails to build without the patch. I think it is because autotools are installed in another directory than xutils-dev so it has to copy xorg-macros.m4 somewhere in the source tree, and for that it needs AC_CONFIG_MACRO_DIR() to be defined.

Here the patch doesn't break the build (on both system and cerbero env) but it's true that it prints a warnings:
aclocal: warning: couldn't open directory 'm4': No such file or directory

Does it work for you if you simply change "m4" directory to "." ?

xclaesse:
Ok, I think I know why it doesn't build for you, AC_CONFIG_MACRO_DIRS needs automake 1.13 that you probably don't have. Made a new pull request that uses the more common AC_CONFIG_MACRO_DIR (no plurial S).

PR #58, opened 2015-06-18
Fix build error when aclocal must copy m4 files

xclaesse:
When ACLOCAL_PATH is set to find m4 files in non standard locations,
aclocal must copy those files in the build tree otherwise automake
won't find them.

Usually AC_CONFIG_MACRO_DIR() is set to m4/ directory but since we
don't ship any m4 file that directory doesn't exist.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -23,6 +23,8 @@ SUBDIRS = include/epoxy src
 
 SUBDIRS += test
 
+ACLOCAL_AMFLAGS = -I .
+
 pkgconfigdir = $(libdir)/pkgconfig
 pkgconfig_DATA = epoxy.pc
 
--- a/configure.ac
+++ b/configure.ac
@@ -27,6 +27,7 @@ AC_INIT([libepoxy],
         [libepoxy])
 AC_CONFIG_SRCDIR([Makefile.am])
 AC_CONFIG_HEADERS([config.h])
+AC_CONFIG_MACRO_DIR([.])
 
 # Initialize Automake
 AM_INIT_AUTOMAKE([foreign dist-bzip2 subdir-objects])


xclaesse:
See pull request #57 for details.

xclaesse:
Any feedback on this ?

anholt:
I see aclocal producing the m4 file in ./aclocal.m4.  Do you maybe have something else in your environment putting them in a nonstandard directory?

xclaesse:
My use case is for cerbero, a tool that builds most of GStreamer dependencies in a prefix (i.e. /opt/cerbero), a bit like GNOME's jhbuild. That tool builds and install aclocal inside the prefix, so the aclocal that will be used to build libepoxy is not system's. That means it won't lookup for m4 files in system directories (i.e. /usr/share/aclocal) but in the prefix (i.e /opt/cerbero/local/aclocal). Since we don't build xorg inside the prefix, it won't find xorg-macros.m4.

To workaround that, we set ACLOCAL_PATH=/usr/local/aclocal in our env to tell aclocal to also lookup there for missing m4 files. The problem now is that aclocal won't copy macros it finds there in the ./aclocal.m4 (dunno why, tbh) but it will want to copy the whole xorg-macros.m4 file into the libepoxy source tree, but it cannot do that if you don't define where to put m4 copies using AC_CONFIG_MACRO_DIR.

I think defining AC_CONFIG_MACRO_DIR is the correct thing to do, libtoolize currently warn you about it:

libtoolize: Consider adding `AC_CONFIG_MACRO_DIR([m4])' to configure.ac and
libtoolize: rerunning libtoolize, to keep the correct libtool macros in-tree.
libtoolize: Consider adding `-I m4' to ACLOCAL_AMFLAGS in Makefile.am.

All other autotools projects I've been checking does it as well, libepoxy is the only exception from what I've seen.

Additionally, since XORG seems to be an optional dependency, it should not require to have xorg-macros.m4 available at build time. The solution would then to ship your own copy of that m4 file.

ebassi:
I have attempted at adding `AC_CONFIG_MACRO_DIR`, but that causes Travis to error out — the default environment for continuous integration is Ubuntu 12.04 LTS. Updating Travis to use 14.04 LTS causes more tests to run, and segfault miserably because 14.04 is basically a stone age environment for Mesa.

Additionally, Epoxy is now slowly migrating to Meson, which Cerbero supports, so I'm not entirely sure tweaking the autotools build is going to be of much help.

PR #61, opened 2015-07-10
Updated "epoxy_is_desktop_gl".

yaronct:
According to the OpenGL ES standrad, "glGetString(GL_VERSION)" Should return a string starting with "OpenGL ES" (This is true for all OpenGL ES versions as of today: 1.0, 1.1., 2.0, 3.0, 3.1). However, PowerVR's OpenGL ES implementation (and perhaps others) don't comply with the standard here. So the only way to distinguish OpenGL ES from Desktop OpenGL in this case is using "epoxy_egl_get_current_gl_context_api()". We can do that because PowerVR's implementation can only be used through EGL.

### DIFF
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -192,6 +192,8 @@ static struct api api = {
 
 static bool library_initialized;
 
+static bool epoxy_current_context_is_glx(void);
+
 #if PLATFORM_HAS_EGL
 static EGLenum
 epoxy_egl_get_current_gl_context_api(void);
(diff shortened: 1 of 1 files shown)


anholt:
I fixed up the style and commit message and pulled it:

commit a7c270f8e2b8dcbf9874632e18a0948255de49d3
Author: Yaron Cohen-Tal <yaronct@gmail.com>
Date:   Thu Jul 9 13:44:52 2015 +0300

    Fix epoxy_egl_get_current_gl_context_api() on non-conformant ES.

PR #64, opened 2015-07-28
src/gen_dispatch.py: Make Generated Code Build Under pre-2013 Visual Studio

fanc999:
Hi,

I am opening this pull request as we would still like to support building GTK+ entirely with Visual Studio, and there are still many people that are not yet able to go up to Visual Studio 2013.

At this point, the main thing that needs to be done is to avoid named initializers, but
instead initialize the structs in old-school C89 way, so the changes needed here are quite small.

The generated code may not look that robust, but since this is generated
code, I think this is not that much an issue; when the Khronos registry gets
updated, all that is needed is that the code gets re-generated, and we have the
items in the right order.

With blessings, thank you!

### DIFF
--- a/src/gen_dispatch.py
+++ b/src/gen_dispatch.py
@@ -659,9 +659,10 @@ class Generator(object):
         assert(offset < 65536)
 
         self.outln('static const uint16_t enum_string_offsets[] = {')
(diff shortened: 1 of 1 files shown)


Pan7:
Works with msvc2010.

PR #69, opened 2015-09-14
Fix for old msvc compilers not having stdbool.h

Pan7:


include/epoxy/gl.h
```diff
@@ -34,7 +34,21 @@
 extern "C" {
 #endif
 
-#include <stdbool.h>
+#if _MSC_VER < 1900
```
ebassi: The indentation is not really keeping up with the rest of the project.

include/epoxy/gl.h
```diff
@@ -34,7 +34,21 @@
...
+	#ifndef __cplusplus
+		#ifndef bool
```
ebassi: That's not how you replace `stdbool.h`.

Instead, you should do something like:

```cpp
#if defined(_MSC_VER) && !defined(__bool_true_false_are_defined) && (_MSC_VER < 1800)
typedef unsigned char bool;
# define false 0
# define true 1
#else
# include <stdbool.h>
#endif
```

PR #70, opened 2015-09-21
dlwrap: add GLIBC_2.4 version to be able to run tests from ARM

blino:
See sysdeps/unix/sysv/linux/arm/libdl.abilist in glibc


yaronct:
@blino: This repository does not seem to b very active recetly, but I've pulled it to [my fork](https://github.com/yaronct/libepoxy), if you're interested.

PR #71, opened 2015-09-22
support newer/larger gl.xml

Nephatrine:
The newest gl.xml from the Khronos registry is too large to fit in the array size. Changed indices from 16-bit to 32-bit and had to change ridiculously enormous string into ridiculously enormous array of single characters as it was over the string size limit.


yaronct:
@Nephatrine: This repository doesn't seem very active lately, but I've pulled it to [my fork](https://github.com/yaronct/libepoxy), in case you're interested.

PR #80, opened 2015-12-20
Allow for X11-less egl builds on linux

gustavoz:
This patch series allows a cross-compiled EGL X11-less build for linux.
It makes sense for embedded targets where you don't want to pull in a big chunk of X11 stack to build gtk3 with wayland for example (via mesa EGL).


jeremyhu:
I think pull request #52 handles this a bit cleaner.  Is there something about that change that doesn't work well for you.

gustavoz:
You still need https://github.com/gustavoz/libepoxy/commit/68e5f1574758240aedfe8653d7aaae62cdb08bf5 since otherwise using non-glx mesa with libepoxy will fail to build properly (or rather libgtk3 with libepoxy in that scenario), like so:

.../sysroot/usr/include/EGL/eglplatform.h:119:22: fatal error: X11/Xlib.h: No such file or directory

Another matter would be to build egl-less glx libepoxy, but that wasn't addressed by me either.

sardemff7:
Any news on these?

gustavoz:
I've already landed all the required patches in buildroot to make this possible, take a look at https://git.busybox.net/buildroot/tree/package/libepoxy
If this hits vanilla libepoxy some day i'll be happy to update/remove the patches.

configure.ac
```diff
@@ -102,6 +102,8 @@ if test x$build_egl = xyes; then
     AC_DEFINE([BUILD_EGL], [1], [build EGL tests])
 fi
 
+AC_SUBST(EGL_CFLAGS)
+
```
sardemff7: No need for that, PKG_CHECK_MODULES does it already.

configure.ac
```diff
@@ -60,7 +60,7 @@ XORG_TESTSET_CFLAG(CWARNFLAGS, [-Wno-int-conversion])
 
 has_znow=yes
 
-case $host_os in
+case $target_os in
```
sardemff7: This is wrong, `$host` is the correct value.
`$build` represents the machine the compiler will run on, `$host` the machine the compiled binaries will run on, and `$target` is the machine the run binary *will produce code for*. See the [Autoconf manual](http://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf.html#Canonicalizing) for reference.
`$target` is only needed for compilers and toolchain utilities.

PR #83, opened 2016-02-26
fixes 'cd error' in autogen.sh

castilma:
If a parent directory of the libepoxy source directory contains a space, 'cd' in line 10 fails.

Note: 'make install' fails, too. I did not fix that


ebassi:
This PR was merged manually in commits 7e9e12f41dd740dff8a81e6b7f32b14242b33cdc and 0f7f16b7320fd19480ad1519346354c8b52662a3.

PR #84, opened 2016-07-28
Fix epoxy_extension_in_string when extension_list is null

manuseiffel:
Fox some X server not supporting any OpenGL feature, glXQueryExtensionsString 
will return NULL and causes the function to fail.

This was verified running an application on macOS while the X server was running
on Windows Xming 7.5.0.10

### DIFF
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -348,7 +348,12 @@ bool
 epoxy_extension_in_string(const char *extension_list, const char *ext)
 {
     const char *ptr = extension_list;
-    int len = strlen(ext);
+    int len;
+    
+    if (!extension_list)
+        return false;
+        
+    len = strlen(ext);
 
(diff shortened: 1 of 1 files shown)


yaronct:
@manuseiffel: The official Epoxy seems to b currently no longer maintained. I try to maintain a [fork](https://github.com/yaronct/libepoxy) of Epoxy - other than being maintained, it also contains a number of bug fixes and improvements, notably it can be built as a static or shared library using cmake, including running tests, and doesn't depend on `eglplatform.h` or `khrplatform.h`. If you're willing to open a pull request to my fork it'll b highly appreciated.

manuseiffel:
@yaronct Done so.

yaronct:
@manuseiffel: thanx!

manuseiffel:
Was fixed in the fork.

PR #86, opened 2016-09-05
nVidia glx.h uses __glx_h__; check it too.

BryanKadzban:
The code used to check this symbol, but commit 689abf4 replaced it with
GLX_H, presumably to work better with either Mesa or Khronos headers.
But nVidia's header use the older include guard.  Add it as another
option.

Fixes the headerguards.c compile test when the system glx.h is from
nVidia's binary drivers.

### DIFF
--- a/include/epoxy/glx.h
+++ b/include/epoxy/glx.h
@@ -39,10 +39,11 @@ extern "C" {
 #include <X11/Xutil.h>
 #include <stdbool.h>
 
-#if defined(GLX_H) || defined(__glxext_h_)
+#if defined(GLX_H) || defined(__glx_h__) || defined(__glxext_h_)
 #error epoxy/glx.h must be included before (or in place of) GL/glx.h
 #else
 #define GLX_H
+#define __glx_h__
 #define __glxext_h_
 #endif
 


yaronct:
@BryanKadzban: This repository of Epoxy seems to b currently no longer maintained. I try to maintain a [fork](https://github.com/yaronct/libepoxy) of Epoxy - other than being maintained, it also contains a number of bug fixes and improvements, notably it can be built as a static or shared library using cmake, including running tests, and doesn't depend on `eglplatform.h` or `khrplatform.h`. If you're willing to open a pull request at my fork it'll b highly appreciated.

BryanKadzban:
"Using cmake"... that's not an improvement.  Requiring that everyone who tries to compile the program-or-library in question also go install some random other tool (...autoconf generates a shell script for a *reason*: /bin/sh and sed are always present, and it doesn't use extensions created in the last ~20 years), dropping lessons learned from porting code across 40-odd years of different unix-like variants plus windows, losing the ability to replace tools like install (see below), losing the configure script's --help flag, that *automatically* tells a package installer what customizations are possible, without them having to go read the cmake config and whatever is used to generate the rest of the build system... all of this is a huge step backwards.

And it also hurts the people writing packages: https://bugs.launchpad.net/kicad/+bug/1603871 / http://public.kitware.com/Bug/view.php?id=10978 -- cmake doesn't search $PATH to run standard tools (like chmod or install), it just has its own code for installing files.  So the package installer can't override behavior without playing LD_PRELOAD tricks (which might not even work; I haven't actually tried).  And that cmake code insists on changing the permissions in places where the user does not have permission to do so, but where the user does have permission to do the installation.  And it's marked wontfix at cmake.

Arrrg.

</screed>

That all said, I found this bug in the released libepoxy-1.3.1 tarball, and I have a copy of this patch locally -- but getting a git tree over to that system to test your version will be a lot more difficult than just the patch that I generated here.  So if it applies to your tree, great, go for it -- but I don't know for sure if it's necessary, and don't have the infrastructure installed yet to try it.

yaronct:
@BryanKadzban: Let me tell u a secret - I don't like cmake more than u do. It does help with Windows though. Anyway, my fork also still supports autotools so if u prefer it u can still use it, so it can't possibly b a step backwards. I'm doing my best to maintain an unmaintained project without dropping any functionality or ABI compatibility. Again, if you'd like to help by creating a PR to my fork, it'll b highly appreciated.

yaronct:
@BryanKadzban: And if u find a better build system that also works on Windows, I'd love to hear about it.

BryanKadzban:
Yeah, I was going to add that but got carried away -- thank you for keeping autotools as an option.  :)

As far as a build system for windows, the best I can come up with is bash-for-windows.  Of course it only supports the most recent version or two (earlier windows releases need cygwin, or mingw/msys, or something like that).  But beyond that -- and the presence of some compiler or other -- autotools *ought* to be able to handle any system differences, I think.  If it doesn't that should be an autotools bug, and they're actually interested in making it portable.

Anyway.

Let me go file a (tentative I guess?) PR -- I won't get a chance to test it for sure but I'll at least dump what I know in there.

yaronct:
@BryanKadzban: Thanx for re-openning the PR in my fork!

I don't mind installing cygwin/msys on windows and indeed with it autotools can handle mingw, however afaik it doesn't know to create Visual Studio solution files for any Visual Studio version, like cmake knows to do.

PR #90, opened 2016-10-18
egl: Fix the query for the current context's API

nwnk:
Binding an API does not change the type of the current context. Even if
it did, EGL 1.5 treats EGL_OPENGL_API and EGL_OPENGLES_API as identical
for this purpose.  If you want to know properties of the current
context, query it.

Signed-off-by: Adam Jackson ajax@redhat.com


chadversary:
Reviewed-by: Chad Versace <chad@kiwitree.net>

PR #92, opened 2016-11-07
build: Force replacing autotools generated files

ebassi:
If the '--force' switch is not passed to autoreconf, the autotools will
try to reuse the generated files committed in the Git repo, which hard
code the names and versions of the tools originally used to generate
them.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -6,7 +6,7 @@ test -z "$srcdir" && srcdir=.
 ORIGDIR=`pwd`
 cd $srcdir
 
-autoreconf -v --install || exit 1
+autoreconf -v -f --install || exit 1
 cd $ORIGDIR || exit $?
 
 if test -z "$NOCONFIGURE"; then


ebassi:
Incidentally: committing autogenerated files to Git is a pretty big "no no". Building from Git without running `autogen.sh` is also fairly broken. Reverting 62e3841 is the appropriate solution.

yaronct:
@ebassi: From `autoreconf`'s manpage:

```
Run  `autoconf'  (and `autoheader', `aclocal', `automake', `autopoint' (formerly `gettextize'), and `libtoolize' where appro‐
       priate) repeatedly to remake the GNU Build System files in specified DIRECTORIES  and  their  subdirectories  (defaulting  to
       `.').

       By default, it only remakes those files that are older than their sources.  If you install new versions of the GNU Build Sys‐
       tem, you can make `autoreconf' remake all of the files by giving it the `--force' option.

...

-f, --force
              consider all files obsolete
```

So, y would we want to make `-f` the default?

ebassi:
In general, because those files you committed to Git encode the name and version of the tool that generated them — and unless everybody who's building libepoxy happens to have your exact same set up, it's going to break the build, e.g.:

```
CDPATH="${ZSH_VERSION+.}:" && cd .. && /bin/sh /ostbuild/source/libepoxy/missing aclocal-1.14 
/ostbuild/source/libepoxy/missing: line 81: aclocal-1.14: command not found
WARNING: 'aclocal-1.14' is missing on your system.
```

Autotools are self-hosting; this is usually perfectly fine for tarballs, but not for Git.

Personally, I think you should revert these commits:

  * 06fd4a271ad88476d8c6a37d49b6a8af5a11d988 (not entirely, but it's a commit that adds `configure`, `config.h.guess` to the Git repo)
  * 62e384129b5b9243634ca4a71108c6f8699a24bd — this is particularly egregious; you should always need to use `autogen.sh` when building from Git; if you don't want to use `autogen.sh` you should build from a dist tarball
  * various commits add generated headers to Git instead of… generating them on demand.

Additionally, autotools can build static libraries just fine — there's really no need to use CMake for that.

yaronct:
@ebassi: The main reason that I added cmake support is for Visual Studio.

I'm not familiar with autotools. What's the difference between git and tarball? Isn't a tarball just a git snapshot? Y wouldn't u have the same problems there?

ebassi:
@yaronct I guessed that you added CMake for Visual Studio. :-)

A dist tarball generated by autotools (using `make dist` or `make distcheck`) is self-hosting: it includes various files needed to build a project with only `make` as a requirement. A Git snapshot is definitely *not* self-hosting because it's assumed that, if you have Git installed, then you also have all the required tools to build a project.

ebassi:
BTW: I just opened PR #93 that just drops the autogenerated files from Git.

yaronct:
@ebassi: Yeah, I know that, so if I just re-reun `autogen.sh` and update all generated files after each commit, so everything should b fine, right? I mean, that's what we do b4 a release, right?

ebassi:
@yaronct Not really, you're not going to be fine. You should run `autogen.sh` every time you need to build from scratch — either you just cloned the repository, or you called `git clean -xdf`. Autobuilders will automatically run `autogen.sh` every time the project changes anyway.

Doing releases with autotools is (usually) a case of:

 * cleaning the repository so that it's pristine
 * rebuild the project from scratch:
  * `NOCONFIGURE=1 ./autogen.sh`
  * `./configure && make && make check`
 * run `make distcheck` to generate the release tarball

People will then be able to use the tarball generated by `make distcheck` to build libepoxy — for instance, packagers of Linux distributions.

yaronct:
@ebassi, @anholt:

Ok, let's make things clear then: The problem u (@ebassi), as a package maintainer, have, isn't specific to a git snapshot - u have exactly the same problem with a release tarball.

Now, I think most Epoxy users aren't package maintainers. This is even true for Epoxy git users. And now I recall that I've made those commits exactly coz such an Epoxy git user complained about having to run "autogen.sh", and that it doesn't work for him coz running it is pretty fragile and depends on the specific versions of the various autotools u have. So I think that for those users having those files committed is convenient, if not to say necessary. The price is that package maintainers have to clean them. It seems to me like a worthy price to pay. But again, I have very little familiarity with those matters, so if there's anything I'm missing plz correct me.

ebassi:
@yaronct It seems there is some miscommunication here, so let's back up from the beginning.

Hi, I'm [Emmanuele](https://bassi.io); I'm a GNOME developer and maintainer, as well as a GTK+ developer. I am not a packager. GTK+ depends on libepoxy (and has for a couple of years), and libepoxy is part of the GNOME [continuous deployment](https://build.gnome.org) and integration pipeline. As such, we try to build GNOME and its dependencies from Git; we haven't been able to build libepoxy for a while — namely, after you pushed commit 62e384129b5b9243634ca4a71108c6f8699a24bd and committed a bunch of autogenerated files to the Git repository. The breakage occurs because the autogenerated files include a reference to the versioned autotools component that generated them, which means that every user of libepoxy now needs to have the exact same version of autotools you used, instead of regenerating those files on their machines. This is a violation of how autotools-based projects should work.

> such an Epoxy git user complained about having to run autogen.sh, and that it doesn't work for him coz running it is pretty fragile and depends on the specific versions of the various autotools u have.

This is wrong, and broken. Letting autotools regenerate those files is *exactly* how autotools are meant to be used.

> So I think that for those users having those files committed is convenient, if not to say necessary.

That is not a price that ought to be paid at all; not only it's entirely inconvenient for anybody who's building and using libepoxy, but it's also broken.

> But again, I have very little familiarity with those matters, so if there's anything I'm missing plz correct me.

The appropriate correction is merging this PR, alongside with PR #93.

I'm entirely happy helping you out with maintaining libepoxy's autotools build system, as you're not familiar with how autotools work.

yaronct:
@ebassi: Thanx for taking the time to explain to me. A random check I've made shows there are projects that do commit autotools generated files to VCS-s (e.g. MPlayer and FreeType), so it's prolly not totally pointless. However I've been convinced that not commiting them makes more sense, as well as using `-f`. Thanx for your contribution!

PR #93, opened 2016-11-07
build: Drop autogenerated autotools files from Git

ebassi:
These files should not be under revision control.

This commit partially reverts:

  * 06fd4a271ad88476d8c6a37d49b6a8af5a11d988
  * 62e384129b5b9243634ca4a71108c6f8699a24bd


anholt:
This is correct.  Autotools-generated files should not be checked into git.

Issue #73, opened 2015-10-21
should have api to safely query GL library existence


jku:
I'm not totally sure where this issue should be fixed but here's the problem:
- GTK+ considers epoxy something they can depend on unconditionally without worrying about the GL implementation
- Installing GTK+ on systems without OpenGL is possible (dependency on epoxy is still there but not really a problem since epoxy is small and doesn't depend on much) 
- running GTK+ applications is a problem: Gdk initialization will try to initialize some visuals for GL and starts by running glXQueryExtension(). This will exit when libGL.so.1 is not present.

So: should epoxy have an api that can be used to check if an implementation is present (without epoxy exiting)? Or should GTK+ do this check on its own?

smcv:
We're hitting this on Debian, including now the installer (!), so I'm going to put a patch together. I spoke to @ebassi on IRC and he said this would be useful to have.

Proposed APIs:

```
bool epoxy_has_{gl,glx,gles1,gles2,egl}_library (void);
```

ebassi:
More something like:

```
  bool epoxy_has_platform_gl (void);
```

which would automatically check for GLX, WGL, AppleGL, or EGL, depending on the platform. Checking for GLES1, 2, 3 is already covered by `epoxy_is_desktop_gl()` and `epoxy_gl_version()`.

smcv:
How do we know whether a non-Mac Unix caller wants GLX or EGL? (On Linux/*BSD/etc. you use GLX for X11, but EGL for Wayland/Mir/misc, right?)

smcv:
To put this another way: what Gtk needs to know is whether glXQueryExtension() is going to work. If you are running under X11, and for some reason you have EGL drivers (Debian: libegl1-mesa) but not GLX (Debian: libgl1-mesa-glx), then a check for "do we have platform GL?" is going to succeed, but  glXQueryExtension() would still make Gtk exit abruptly.

yaronct:
@smcv, you're right, we can't know which API is used to create an opengl context. So I think we have to b queried for something specific like "do we have api XXX?".

What then would b your requirements for gtk as for "XXX"? GLX? OpenGL?

smcv:
At the moment GTK itself only needs to know whether `glXQueryExtension()` is going to crash the process, so `epoxy_has_glx()` or similar. Expected result: if there is `libGL.so.1` and we can dlopen it and we can get version information or whatever, return true; if not, return false; under no circumstances, call exit() :-)

[...]

PR #98, opened 2017-01-27
Platform detection

ebassi:
Adds functions for detecting, at run time, whether EGL and/or GLX are available on the current platform.

Fixes issue #73.


ebassi:
This PR is based off of the `ebassi-next` branch — i.e. PR #97.

src/dispatch_glx.c
```diff
@@ -109,3 +111,27 @@ epoxy_has_glx_extension(Display *dpy, int screen, const char *ext)
...
+     * code but already figured out how to load GL symbols, then we
+     * simply inspect the address space.
+     *
+     * We explicitly don't use the epoxy wrappers, here, to avoid
+     * unintended side effects within our own layer of indirection.
+     */
+    Bool (* pf_glXQueryExtension) (Display *, int *, int *);
+    int error_base, event_base;
+
+    pf_glXQueryExtension = epoxy_glx_dlsym("glXQueryExtensions");
```
smcv: Don't you want to be calling a variant of `epoxy_glx_dlsym` that passes `false` for the `exit_on_fail` parameter?

I think this implementation would just exit 1 in an environment with no libGL.so.1, where we want it to return false.

Issue #43, opened 2015-03-17
SyntaxError: invalid syntax during "make"


felibb:
Hi,

I am trying to compile libepoxy on AWS Linux, and I get this error during "make":

``` shell
/bin/mkdir -p ../include/epoxy
  GEN      gl_generated_dispatch.c
  File "./gen_dispatch.py", line 67
    'glBegin',
             ^
SyntaxError: invalid syntax
make[2]: *** [gl_generated_dispatch.c] Error 1
make[2]: Leaving directory `/opt/libepoxy-1.2/src'
make[1]: *** [all-recursive] Error 1
make[1]: Leaving directory `/opt/libepoxy-1.2'
make: *** [all] Error 2
```

autogen.sh ran without errors. Any help is appreciated.

anholt:
It looks like maybe you've got a python error, except that it looks truncated somehow?  It doesn't make any sense to me.

nwnk:
It's not truncated. The bug is that's not legal syntax in python 2.6.

nwnk:
Something like this gets it to build on RHEL6.7, though obviously can't be merged as-is since it drops the wgl and gles bits.

```
diff -up libepoxy-1.2/src/gen_dispatch.py.jx libepoxy-1.2/src/gen_dispatch.py
--- libepoxy-1.2/src/gen_dispatch.py.jx	2015-10-28 11:23:44.441741032 -0400
+++ libepoxy-1.2/src/gen_dispatch.py	2015-10-28 11:24:03.604898390 -0400
@@ -63,14 +63,7 @@ class GLFunction(object):
         # These are functions with hand-written wrapper code in
         # dispatch_common.c.  Their dispatch entries are replaced with
         # non-public symbols with a "_unwrapped" suffix.
-        wrapped_functions = {
-            'glBegin',
-            'glEnd',
-            'wglMakeCurrent',
-            'wglMakeContextCurrentEXT',
-            'wglMakeContextCurrentARB',
-            'wglMakeAssociatedContextCurrentAMD',
-        }
+        wrapped_functions = [ 'glBegin', 'glEnd' ]
 
         if name in wrapped_functions:
             self.wrapped_name = name + '_unwrapped'
@@ -204,7 +197,7 @@ class Generator(object):
 
     def parse_typedefs(self, reg):
         for t in reg.findall('types/type'):
-            if 'name' in t.attrib and t.attrib['name'] not in {'GLhandleARB'}:
[...]

PR #100, opened 2017-01-28
meson: Don't call find_program objects with python

nirbheek:
On UNIX-like OSes, the OS will read the shebang and use the correct interpreter, and on Windows, Meson will read the shebang and use the correct interpreter.

Adding it manually will cause python to try to interpret python.exe

With this change, libepoxy builds and the tests pass with MSVC on Windows for me.

### DIFF
--- a/include/epoxy/meson.build
+++ b/include/epoxy/meson.build
@@ -4,7 +4,6 @@ gl_generated = custom_target('gl_generated.h',
                                'gl_generated.h',
                              ],
                              command: [
-                               python,
                                gen_dispatch_py,
                                '--header',
                                '--no-source',
@@ -24,7 +23,6 @@ if build_egl
                                   'egl_generated.h',
                                 ],
                                 command: [
-                                  python,
                                   gen_dispatch_py,
                                   '--header',
                                   '--no-source',
@@ -44,7 +42,6 @@ if build_glx
                                   'glx_generated.h',
                                 ],
                                 command: [
-                                  python,
                                   gen_dispatch_py,
                                   '--header',
                                   '--no-source',
@@ -64,7 +61,6 @@ if build_wgl
                                   'wgl_generated.h',
                                 ],
                                 command: [
-                                  python,
                                   gen_dispatch_py,
                                   '--header',
                                   '--no-source',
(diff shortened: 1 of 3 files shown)


ebassi:
Interesting; I did add the explicit Python call because of MSYS2 build errors. Can you confirm that it still builds with Meson 0.37 on MSYS2?

nirbheek:
I tested this on native Windows and on MSYS, but not on MSYS2 since I don't use it. However this is how we do things in GStreamer too and other people have tested that it works there. Will confirm and report back though!

You might also see a PR adding AppVeyor CI support to build with MSVC and MinGW. :)

nirbheek:
So it doesn't work on MSYS2, because there Python 3.5 is called `python3.exe` and `python3.5.exe` but not `python.exe`. So `#!/usr/bin/env python` in the `gen_dispatch.py` can't find `python`.

The fix is to use `#!/usr/bin/env python3` as the shebang, but that will not work on native Windows with Meson 0.37.1 since Python 3 installs itself as `python.exe` with the Windows installer. However, Meson 0.38 will correctly use [`sys.executable`](sys.executable) on Windows when it encounters `python3` in a shebang, so it should be ok to do that in the long run.

nirbheek:
I will note that this same problem also occurs on Linux distros since no one ships Python 3 as `python.exe`. Essentially Python 2 is required on all platforms for building libepoxy, and if you install Python 2 inside MSYS2 it builds just fine.

ebassi:
Ideally, no: Epoxy should build with Python 3.

The only reason why Python 2 may be used is because of older, enterprise-y environments — see issue #43.

I guess we need to change `gen_dispatch.py` to take the right version of Python in the shebang line.

nirbheek:
I was hoping there's a way to avoid that, but such is life. ;)

ebassi:
Okay, so I kind of changed my mind, given that Epoxy has to build on platforms without Python 3, which means we're going to keep `gen_dispatch.py` compatible with Python 2 for the foreseeable future.

I'll merge this PR, and then update the documentation to say that we require Python 2 for building Epoxy.

Issue #108, opened 2017-03-03
meson: add library versioning?


tschoonj:
Hi,

I maintain the `libepoxy` formula in Homebrew. After yesterday's release I decided to give the new meson build system a [try](https://github.com/Homebrew/homebrew-core/pull/10566), but noticed that the produced dylib is not versioned. This is a bit of a problem, as it breaks our gtk+3 binaries, and all packages that depend on it. 

The configure script based install produces:

````
/usr/local/Cellar/libepoxy/1.4.1/lib/libepoxy.0.dylib
/usr/local/Cellar/libepoxy/1.4.1/lib/libepoxy.dylib
````

with libepoxy.dylib being a symlink to libepoxy.0.dylib

and with the meson installation:

````
/usr/local/Cellar/libepoxy/1.4.1/lib/libepoxy.dylib
````

Since the [meson docs](https://github.com/mesonbuild/meson/wiki/Build-targets) mention support for versioned libraries, I wondered if it would be useful to use this feature? To me it makes sense that both installation types produce the exact same files.

Best,

Tom

ebassi:
Hi; thanks for the bug report.

We currently set the [`version`](https://github.com/anholt/libepoxy/blob/master/src/meson.build#L82) field of the shared library to `0.0.0`, as libepoxy has not been using a versioned shared library since its inception - see also: #1 for a bug about adding per-symbol versioning with GNU libc to counter this.

On Linux, Epoxy generates a `libepoxy.so.0.0.0` file, with a link to `libepoxy.so` and `libepoxy.so.0`.

I'm going to ask the Meson developers how to generate a `dylib` file with the same layout.

ebassi:
@tschoonj Could you try adding a `soversion: 0` field after the `version` one in the file I linked above?

ebassi:
It turns out that Meson intentionally strips the version from the filename; see mesonbuild/meson#680 — which was introduced because of the linker's inability to find the unversioned shared library, i.e. `-lfoo` and `-lfoo.0` are different.

The [Apple documentation](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryDesignGuidelines.html#//apple_ref/doc/uid/TP40002013-SW20) seems to suggest that the autotools interpretation is correct.

[...]

PR #109, opened 2017-03-07
Add explicit version flags for macOS builds

ebassi:
Autotools automatically adds version related linker flags on macOS that
are not related to the version information. Since we want to maintain
binary compatibility, we should do the same when building on macOS.

Fixes: #108

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -51,10 +51,14 @@ foreach h: headers
   epoxy_headers += join_paths(meson.source_root(), 'include/epoxy/@0@'.format(h))
 endforeach
 
+common_ldflags = []
 if cc.get_id() == 'gcc'
   common_ldflags = [ '-Wl,-Bsymbolic', '-Wl,-z,relro', '-Wl,-z,now', ]
-else
-  common_ldflags = []
+endif
+
+# Maintain compatibility with autotools; see: https://github.com/anholt/libepoxy/issues/108
+if host_system == 'darwin'
+  common_ldflags += [ '-compatibility_version=1', '-current_version=1.0', ]
 endif
 
 epoxy_deps = [ dl_dep, ]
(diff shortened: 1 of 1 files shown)


src/meson.build
```diff
@@ -51,10 +51,14 @@ foreach h: headers
   epoxy_headers += join_paths(meson.source_root(), 'include/epoxy/@0@'.format(h))
 endforeach
 
+common_ldflags = []
 if cc.get_id() == 'gcc'
   common_ldflags = [ '-Wl,-Bsymbolic', '-Wl,-z,relro', '-Wl,-z,now', ]
```
nirbheek: Btw, I wanted to mention this elsewhere—aren't these options also supported when clang/icc is the compiler since these are passed to ld anyway? So this should be `cc.get_id() != 'msvc'`?

ebassi: I'm not entirely sure; I'd probably switch that to:

  `if host_machine == 'linux'`

instead.

src/meson.build
```diff
@@ -51,10 +51,14 @@ foreach h: headers
...
-else
-  common_ldflags = []
+endif
+
+# Maintain compatibility with autotools; see: https://github.com/anholt/libepoxy/issues/108
+if host_system == 'osx'
```
nirbheek: This should be 'darwin'.

src/meson.build
```diff
@@ -51,10 +51,14 @@ foreach h: headers
...
+  common_ldflags += [ '-compatibility_version 1', '-current_version 1.0', ]
```
nirbheek: I think these have to be `=1` or four list elements since they are four separate arguments.

src/meson.build
nirbheek: You're right, I don't think these options are supported on [macOS ld](http://www.manpages.info/macosx/ld.1.html).

ebassi:
Thanks for the review, @nirbheek.

PR #112, opened 2017-03-14
Fix build on MSVC

accel2k:
MSVC 2013 failed to build libepoxy due to error C2054: expected '(' to follow 'inline'.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -159,6 +159,14 @@ if libtype == 'shared'
   endif
 endif
 
+# The inline keyword is available only for C++ in MSVC.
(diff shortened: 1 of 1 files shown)


ebassi:
Hi! Thanks for your patch!

Would it be possible for you to leave a comment, either inside the build system or as a commit message, detailing the fix? Something simple, like what you wrote in the message of the PR would be perfect. This would avoid me doing dumb stuff six months down the line.

Thanks again!

accel2k:
Hi!
I added comment as you asked.

ebassi:
Thanks!

PR #113, opened 2017-03-15
Forward EGL cflags into epoxy.pc

gustavoz:
When building mesa egl without x11 and gles2 the headers need a
MESA_EGL_NO_X11_HEADERS define to avoid including X11 headers.
Forward EGL CFLAGS into epoxy.pc to avoid build failures when linking
against libepoxy on this scenario.

Signed-off-by: Gustavo Zacarias <gustavo@zacarias.com.ar>

### DIFF
--- a/epoxy.pc.in
+++ b/epoxy.pc.in
@@ -10,6 +10,6 @@ epoxy_has_wgl=@epoxy_has_wgl@
 Name: epoxy
 Description: epoxy GL dispatch Library
 Version: @PACKAGE_VERSION@
-Cflags: -I${includedir}
+Cflags: -I${includedir} @EGL_CFLAGS@
 Libs: -L${libdir} -lepoxy
 Libs.private: @DLOPEN_LIBS@


anholt:
I don't think Mesa cflags should be in our .pc.  But maybe we need to #define this before including eglplatform.h, which is the only thing that epoxy includes from Mesa.

(I've thought before about providing an eglplatform.h from Epoxy as well, so there's no direct exposure of system GL/EGL headers to our users)

ebassi:
I agree with @anholt: leaking this definition inside the epoxy pkg-config file is not the right approach.

If you're building Epoxy on a system without X11 and Mesa, then you should already building Epoxy without GLX support; if that happens, you can detect it in your own code by checking the `epoxy_has_glx` variable inside the pkg-config, and add the definition yourself, as a workaround.

Epoxy can make this a bit easier, and define `MESA_EGL_NO_X11_HEADERS` when including `eglplatform.h`, since it knows when it's being built without GLX.

gustavoz:
I'm not an expert with *GL* so bare with me.
Isn't the point of libepoxy for people to abstract themselves from the little implementation details?
I think @anholt is spot on with the #define and packages that use libepoxy shouldn't deal with that nuissance.
Alternatively providing a santized eglplatform.h sounds good as well.
(i'm normally maintaining/using this in buildroot which is a build-from-scratch tool targetting embedded devices, so not dealing with code on my own).

gustavoz:
@anholt @ebassi  Another question before i go patch happy: is EGL a hard requirement when using just GLX? (with X11 obviously)
I previously patched this away and worked fine.

clopez:
This is related to issue #139
But I think the right fix is to add to the pkg-config file:
* if support for glx and egl are enabled:
```
Requires.private: glx egl
```
* if only support for glx is enabled:
```
Requires.private: glx
```
* if only support for egl is enabled:
```
Requires.private: egl
```

ebassi:
See PR #153 for an alternative approach

clopez:
> See PR #153 for an alternative approach

That is just a workaround for Mesa, its not the right fix IMHO.
This issue happens not only when Mesa is built without X11 support, but also with other EGL libraries like the I.MX6 Vivante one that define several flags on its pkg-config that should be defined when including it.

ebassi:
@clopez We can add more defines in the same header. I really don't think it should be part of the pkg-config file, because that implies I would not be able to take a libepoxy build done on a platform using Mesa and use it on another one.

clopez:
> @clopez We can add more defines in the same header. I really don't think it should be part of the pkg-config file, because that implies I would not be able to take a libepoxy build done on a platform using Mesa and use it on another one.

I think you can do that. All that needs to be done is to add this to the epoxy pkg-config file.

```
Requires.private: egl
```

And pkg-config will at run-time automatically get the cflags from the EGL pkg-config library of the current system and append that to the provided CFLAGS by epoxy. This will allow the build to complete on the current system because those flags are needed by ```EGL/eglplatform.h``` that epoxy includes from ```egl_generated.h```. But the resulting epoxy binary should still work with another EGL libraries, as this extra CFLAGS are just compile-time definitions (macros) needed by eglplatform.h to not trigger an ``` #error something``` when the header  ```EGL/eglplatform.h``` is included

clopez:
BTW: requires.private doesn't change the flags passed to the linker (the value of ``` pkg-config --libs epoxy```), it only changes the compile-time definitions (value of ```pkg-config --cflags epoxy```). So I think this can't affect the resulting epoxy binaries in any case because the extra macros defined by the system egl.pc should be a no-op when building epoxy itself.

ebassi:
@clopez You convinced me; I ended up adding `gl` and `egl`, if found and configured, to the `Requires.private` field of the pkg-config file.

PR #114, opened 2017-03-19
test: Fix dlwrap on aarch64

Millak:
dlsym on aarch64 has version GLIBC_2.17.


ebassi:
Thanks for testing, and for the patch!

Issue #115, opened 2017-03-27
Cannot be used as a meson subproject (wrap)


dcbaker:
`meson.source_root()` will always be the toplevel source root, so if libepoxy is used as a wrap the generation of epoxy_headers will fail since it will look in the for `include/epoxy` in the parent project, not in epoxy itself.

ebassi:
Indeed, we use `meson.source_root()` twice:

 - once to find the `gen_dispatch.py` script, to work around a Meson bug that was already fixed a couple of releases ago
 - once to construct a full path to the headers in order to generate appropriate dependencies, as we refer to those files from different directories

Both should go away regardless.

ebassi:
@dcbaker I've opened a PR that should fix the issue; testing is appreciated.

PR #116, opened 2017-03-28
Remove uses of meson.source_root()

ebassi:
See issue: #115


dcbaker:
This does fix the issue for me. Thanks! :)

PR #119, opened 2017-05-08
Make EGL support optional

tpetazzoni:
It is perfectly possible to build Mesa3D with just OpenGL support, and
use with GLX in X.org, without having EGL/OpenGLES support.

However, libepoxy currently unconditionally requires EGL support in its
configure.ac, which causes a build failure when Mesa3D only provides
full OpenGL support:

checking for EGL... no
configure: error: Package requirements (egl) were not met:

Package egl was not found in the pkg-config search path.
Perhaps you should add the directory containing `egl.pc'
to the PKG_CONFIG_PATH environment variable
Package 'egl', required by 'world', not found

This commit fixes that by:

 - Adjusting the configure.ac to add a --{enable,disable}-egl option
   handled in the exact same way as --{enable,disable}-glx

 - Adjusting src/dispatch_common.h to define PLATFORM_HAS_EGL correctly,
   which allows to not include any EGL related header file if EGL
   support is not enabled.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>


tpetazzoni:
Note that I am far from being familiar with libepoxy, so please do a very careful review.

ebassi:
Would it be possible for you to add the changes to the Meson build system as well?

tpetazzoni:
Sure, I'll submit a separate pull request.

PR #123, opened 2017-05-17
Make EGL support optional

tpetazzoni:
It is perfectly possible to build Mesa3D with just OpenGL support, and
use with GLX in X.org, without having EGL/OpenGLES support.

However, libepoxy currently unconditionally requires EGL support in its
configure.ac, which causes a build failure when Mesa3D only provides
full OpenGL support:

checking for EGL... no
configure: error: Package requirements (egl) were not met:

Package egl was not found in the pkg-config search path.
Perhaps you should add the directory containing `egl.pc'
to the PKG_CONFIG_PATH environment variable
Package 'egl', required by 'world', not found

This commit fixes that by:

 - Adjusting the configure.ac to add a --{enable,disable}-egl option
   handled in the exact same way as --{enable,disable}-glx

 - Adjusting the meson build logic in the same way.

 - Adjusting src/dispatch_common.h to define PLATFORM_HAS_EGL correctly,
   which allows to not include any EGL related header file if EGL
   support is not enabled.

Signed-off-by: Thomas Petazzoni <thomas.petazzoni@free-electrons.com>


configure.ac
```diff
@@ -88,11 +88,35 @@ AS_CASE([$enable_glx],
...
+              [AC_HELP_STRING([--enable-egl=@<:@auto,yes,no@:>@], [Enable EGL support @<:@default=auto@:>@])],
+              [enable_egl=$enableval],
+              [enable_egl=auto])
+
+AS_CASE([$enable_egl],
+  [auto], [
+    AS_CASE([$host_os],
+            [mingw*], [build_egl=no],
+            [darwin*], [build_egl=no],
+            [android*], [build_egl=no],
```
ebassi: I'm pretty sure Android *only* has EGL, so it should not be listed there.

meson.build
```diff
@@ -51,26 +51,41 @@ elif enable_glx == 'no'
...
 endif
 
+enable_egl = get_option('enable-egl')
+if enable_egl == 'auto'
+  if host_system == 'windows'
+    build_egl = false
+  elif host_system == 'darwin'
+    build_egl = false
+  elif host_system == 'android'
+    build_egl = false
```
ebassi: Same as above: Android should have EGL built by default.

ebassi:
As an addition, it would be great if your pull request also modified the `.travis.yml` file to include a build of Epoxy with `-Denable-egl=false`, just like we test the build without GLX. This would allow us to test the newly added configuration switch and ensure it's not breaking.

ebassi:
Thanks for working on this! I merged your branch and fixed the issues above.

tpetazzoni:
Thanks!

PR #124, opened 2017-05-22
Add support for building on Haiku.

return:
Haiku is a operating system that targets personal computing and is inspired by BeOS. This commit disables building glx on Haiku, allowing it to build libepoxy successfully. I also tested the build on Haiku and it passes all tests.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -42,6 +42,8 @@ if enable_glx == 'auto'
     build_glx = false
   elif host_system == 'android'
     build_glx = false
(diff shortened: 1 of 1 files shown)


ebassi:
Thanks! Do you have a CI environment to test changes to Epoxy so that we don't accidentally break the build?

return:
@ebassi Not yet, I haven't set one up for Haiku. But I'll see if it is possible to do so.

PR #128, opened 2017-06-12
meson.build: fix macOS common_ldflags

tschoonj:
Just saw this happening on macOS:

````
[9/13] Linking target src/libepoxy.0.dylib
FAILED: src/libepoxy.0.dylib 
clang  -o src/libepoxy.0.dylib 'src/epoxy@sha/src_gl_generated_dispatch.c.o' 'src/epoxy@sha/dispatch_common.c.o' '-shared' '-install_name' '/private/tmp/libepoxy-20170612-40467-rny9re/libepoxy-1.4.3/build/src/libepoxy.dylib' '-compatibility_version=1' '-current_version=1.0' '-ldl' '-Wl,-rpath,/private/tmp/libepoxy-20170612-40467-rny9re/libepoxy-1.4.3/build/src'  
ld: malformed 32-bit x.y.z version number: =1
clang: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.

````

Fix was simply replacing the equal signs with spaces.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -59,7 +59,7 @@ endif
 
 # Maintain compatibility with autotools; see: https://github.com/anholt/libepoxy/issues/108
 if host_system == 'darwin'
-  common_ldflags += [ '-compatibility_version=1', '-current_version=1.0', ]
+  common_ldflags += [ '-compatibility_version 1', '-current_version 1.0', ]
 endif
 
 epoxy_deps = [ dl_dep, ]

### CODE REVIEW

File: src/meson.build [hunk 1]
```diff
@@ -59,7 +59,7 @@ endif
 
 # Maintain compatibility with autotools; see: https://github.com/anholt/libepoxy/issues/108
 if host_system == 'darwin'
-  common_ldflags += [ '-compatibility_version=1', '-current_version=1.0', ]
+  common_ldflags += [ '-compatibility_version 1', '-current_version 1.0', ]
```
ebassi: Shouldn't this be:

```
[ '-compatibility_version', '1', '-current_version', '1.0' ]
```

tschoonj: The patch works as it is now already, but I will gladly make the change if you prefer it this way

ebassi: I'm generally worried about the quoting rules, but since you've tested this, and it's macOS only, I'm going to just go with it.

PR #129, opened 2017-06-16
dispatch: Learn about glvnd

nwnk:
glvnd has the rather nice property that GLX and (desktop) GL are
available in separate libraries. As an example this would allow an EGL
app to use desktop GL without any X11 libraries. Fix up our dlopens to
prefer the glvnd libraries if available and fall back to the old ABI if
not.

Signed-off-by: Adam Jackson <ajax@redhat.com>

### DIFF
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -178,6 +178,7 @@
 #elif defined(ANDROID)
 #define GLX_LIB "libGLESv2.so"
 #else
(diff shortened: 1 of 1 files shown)


anholt:
Looks good to me.  I'm guessing ebassi will merge, though :)

ebassi:
Indeed, it looks good.

PR #133, opened 2017-07-12
Try even harder to not load GLX

nwnk:
This doesn't _quite_ eliminate X11 libraries from EGL applications, since
Mesa's libEGL has an X11 backend. But it does eliminate libGL and libGLX.


ebassi:
Thanks, it looks good.

There are a few whitespace snafus — probably tabs vs spaces — but as we don't have a coding style document, I'm not in a position to ask for a change.

PR #136, opened 2017-08-24
Expose epoxy_extension_in_string() in public headers

Lyude:
The EGL_EXT_device_query extension introduces the
eglQueryDeviceStringEXT() function, which can be used with an
EGLDeviceEXT enumeration to receive an OpenGL extension string
containing all of the device extensions supported by it. From the
EGL_EXT_device_query spec's amendments to section "3.2 Devices" after
"3.1 Errors":

    const char *eglQueryDeviceStringEXT(EGLDeviceEXT device,
    				    EGLint name);

    returns a pointer to a static, zero-terminated string describing
    some aspect of the specified EGLDeviceEXT.  <name> must be
    EGL_EXTENSIONS.

Since OpenGL extension parsing is rather simple, and we could always run
into additional cases similar to this one in the future, we expose the
shared epoxy_extension_in_string() function in libepoxy's public headers
so that users can save themselves the hassle of having to write their
own extension parser.

Signed-off-by: Lyude Paul <thatslyude@gmail.com>

### DIFF
--- a/include/epoxy/common.h
+++ b/include/epoxy/common.h
@@ -53,4 +53,11 @@ typedef unsigned char bool;
 # include <stdbool.h>
 #endif
 
+EPOXY_BEGIN_DECLS
+
+EPOXY_PUBLIC bool epoxy_extension_in_string(const char *extension_list,
+                                            const char *ext);
+
+EPOXY_END_DECLS
+
 #endif /* EPOXY_COMMON_H */
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -446,6 +446,24 @@ epoxy_conservative_gl_version(void)
     return epoxy_internal_gl_version(100);
 }
 
+/**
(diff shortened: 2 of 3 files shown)


ebassi:
How would this differ from `epoxy_has_egl_extension()`? Outside of the marginal performance win of getting the static string once, I mean…

nwnk:
`epoxy_has_egl_extension` parses either the extension string of the display or of no display (effectively, of the client library). `eglQueryDeviceStringEXT` adds another kind of extension string, hanging off a different object, a device.

Lyude:
Yep, that's the gist of it. There aren't really any performance reasons for this; I just figured if we've already got library helpers for the rest of the GL/EGL/etc. APIs we might as well have one for edge cases in the spec like this.

ebassi:
Sounds fair, then.

PR #141, opened 2017-10-13
meson: Define visibility flags for static builds

dcbaker:
It's used unconditionally, but defined only for shared builds. For
static builds define it but don't populate it.


ebassi:
Nice catch, thanks!

PR #143, opened 2017-10-19
Fix printf family usage

ikeydoherty:
- Fix use of `asprintf` to be correct and silence compiler warnings
 - Only use `fprintf` when actually needed, otherwise use `fputs` (avoid unnecessary allocations)


ebassi:
Thanks!

ikeydoherty:
np :)

PR #144, opened 2017-10-20
change hardcoded library names to less hardcoded names.

bramoosterhuis:
Sharing is caring, after an upgrade of WebKit I needed to use libepoxy on several platforms that are not based on Android, which didn't have e.g. libEGL.so.1 and I wasn't allowed to introduce extra symbolic links. So I thought there must be a way to specify the graphics library names from a build system level.  This is maybe not the cleanest way, but this way you can add the names using defines at build time.  e.g. in my buildroot makefile: 

ifeq ($(BR2_PACKAGE_PLATFORM_X),y)
LIBEPOXY_CONF_OPTS += \
	CFLAGS='$(TARGET_CFLAGS) \
			-DGLX_LIB_NAME="libGLESv2.so" \
			-DEGL_LIB_NAME="libEGL.so" \
			-DGLES1_LIB_NAME="libGLESv1_CM.so" \
			-DGLES2_LIB_NAME="libGLESv2.so"'
else
LIBEPOXY_MAKE_OPTS += \
	CFLAGS='$(TARGET_CFLAGS) \
            -DGLX_LIB_NAME="libGL.so.1" \
            -DEGL_LIB_NAME="libEGL.so.1" \
            -DGLES1_LIB_NAME="libGLESv1_CM.so.1" \
            -DGLES2_LIB_NAME="libGLESv2.so.2"'
endif

### DIFF
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -177,7 +177,7 @@
 #define str(a) #a
 
 #ifndef GLX_LIB_NAME
-#error "GLX_LIB_NAME not defined"
+#define GLX_LIB "libGL.so.1"
 #else
 #define GLX_LIB    xstr(GLX_LIB_NAME)
 #endif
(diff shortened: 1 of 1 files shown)


ebassi:
Not having fallbacks in place is clearly not going to work — as the CI shows.

Additionally, this has to work with both Autotools (on their way out) and Meson/Ninja.

ebassi:
This is still not going to work for Meson builds.

Ideally, I'd rather have a configure-time option instead of adding random `#define` pre-processor directives to the compiler flags.

Something like:

```
$ ./configure --with-opengl-lib=/some/weird/prefix/lib/libGL.so.42
```

or:

```
$ meson -Dopengl_lib=/some/weird/prefix/lib/libGL.so.42 _build
```

PR #151, opened 2018-01-18
Check __ANDROID__ define when guarding Android code

rib:
Both gcc and clang define `__ANDROID__` but not `ANDROID` when targeting Android so
at least with the Meson build, and an NDK r16 toolchain, testing on
Android failed without this change.

Afaik `ANDROID` gets defined by `ndk-build` which is one reason to see code check that instead, but I'm not sure what ensured it was defined for whoever originally added the Android support to libepoxy. Maybe they used an external Android.mk based build or just added `-DANDROID` to their cflags, not sure.


ebassi:
Thanks for testing on Android!

PR #152, opened 2018-01-23
Guard OPENGL_LIB dependent code

rib:
This guards the epoxy_gl_dlsym code that depends on the OPENGL_LIB
define. This fixes building for Android or Windows.

### DIFF
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -648,7 +648,7 @@ epoxy_gl_dlsym(const char *name)
     return do_dlsym(&api.gl_handle,
                     "/System/Library/Frameworks/OpenGL.framework/Versions/Current/OpenGL",
                     name, true);
-#else
+#elif defined(OPENGL_LIB)
     void *sym;
 
     if (!api.gl_handle)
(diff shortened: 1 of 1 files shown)


src/dispatch_common.c
```diff
@@ -648,7 +648,7 @@ epoxy_gl_dlsym(const char *name)
     return do_dlsym(&api.gl_handle,
                     "/System/Library/Frameworks/OpenGL.framework/Versions/Current/OpenGL",
                     name, true);
-#else
+#elif defined(OPENGL_LIB)
```
ebassi: I'm not sure this is correct.

The `#if defined(OPENGL_LIB) ... #endif` block should only be used around the:

```cpp
    if (!api.gl_handle)
        get_dlopen_handle(&api.gl_handle, OPENGL_LIB, false);
```

block. The following block will call `dlsym()` if the `gl_handle` is set, and if it isn't, it'll do a `dlopen()`.

The only difference between the Android and Linux paths is that on Linux there's an `OPENGL_LIB` define in the latter but not in the former.

ebassi: The commit message is also not really accurate: the Windows conditional path already exists.

ebassi:
I ended up pushing a possible fix; please, feel free to re-open if the build still fails.

PR #153, opened 2018-02-07
Define MESA_EGL_NO_X11_HEADERS before including eglplatform.h when X1…

kraj:
…1 is disabled

This helps in compiling when GL implementations e.g. mesa use
this define to exclude X11 headers

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/dispatch_common.h
+++ b/src/dispatch_common.h
@@ -46,6 +46,14 @@
 #include "epoxy/glx.h"
 #endif
 #if PLATFORM_HAS_EGL
+#if !PLATFORM_HAS_GLX
+/* GL implementations e.g. mesa and mali use this define to exclude X11
+ * headers, here its known that GLX is not enabled therefore define it
+ * so subsequent including eglplatform.h can avoid needing X11 headers
+ * when building for non X11 system.
+ */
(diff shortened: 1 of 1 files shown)


src/dispatch_common.h
```diff
@@ -46,6 +46,9 @@
 #include "epoxy/glx.h"
 #endif
 #if PLATFORM_HAS_EGL
+#if !PLATFORM_HAS_GLX
+#define MESA_EGL_NO_X11_HEADERS 1
```
ebassi: It would be nice to leave a comment, here, explaining what this define is.

kraj:
@ebassi thanks done in V2

ebassi:
Mmh, thinking a bit more about this, I don't think it's right at all.

Disabling GLX does not imply disabling X11 support on EGL; those are two very orthogonal things.

This needs a separate build option, or it requires that your build system injects `-DMESA_EGL_NO_X11_HEADERS` inside the `CFLAGS` variable when building on a system without X11.

ebassi:
I ended up pushing a different fix, which allows you to completely disable X11 support when building Epoxy.

PR #155, opened 2018-02-22
test: Fix respecting $(DLOPEN_LIBS)

mgorny:
Fix two tests that are linking to -ldl directly to use DLOPEN_LIBS. This
fixes failing test build on Gentoo/FreeBSD.


ebassi:
Thanks!

PR #156, opened 2018-02-28
meson: do not add gl to pkg-config file on macOS

tschoonj:
Even though meson will find the dependency gl on macOS, this does not mean that
there is a pkg-config file for it, as meson does not use pkg-config to
establish its presence. It should therefore not be added to
the libepoxy pkg-config file as a (private) requirement.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -93,10 +93,10 @@ epoxy_has_wgl = build_wgl ? '1' : '0'
 # not needed when building Epoxy; we do want to add them to the generated
 # pkg-config file, for consumers of Epoxy
 gl_reqs = []
-if gl_dep.found()
+if gl_dep.found() and gl_dep.type_name() == 'pkgconfig'
   gl_reqs += 'gl'
 endif
-if build_egl and egl_dep.found()
+if build_egl and egl_dep.found() and egl_dep.type_name() == 'pkgconfig'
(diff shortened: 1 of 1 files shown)


tschoonj:
Found that this creates trouble building anything that depends on `libepoxy` on macOS. I can confirm that this patch fixes the problem.

src/meson.build
```diff
@@ -93,7 +93,7 @@ epoxy_has_wgl = build_wgl ? '1' : '0'
 # not needed when building Epoxy; we do want to add them to the generated
 # pkg-config file, for consumers of Epoxy
 gl_reqs = []
-if gl_dep.found()
+if gl_dep.found() and host_system != 'darwin'
```
ebassi: I'm pretty sure this will fail on Windows as well.

The appropriate solution would be to check the type of the `gl_dep` object and adding it only if it's a pkg-config dependency, e.g.

```
if gl_dep.found() and gl_dep.type_name() == 'pkgconfig'
  gl_reqs += 'gl'
endif
```

Same goes for the EGL dependency:

```
if build_egl and egl_dep.found() and egl_dep.type_name() == 'pkgconfig'
  gl_reqs += 'egl'
endif
```

tschoonj:
Done!

Tested on macOS.

PR #158, opened 2018-03-01
Test tweaks

rossburton:
Two patches:

One to add an option to disable the test suite entirely for eg distro builders that won't run it

The other checks if dlvsym is present as that is a glibc-specific function and the tests fail to build on musl.  Note that there's an existing "if not apple" check in all the functions which use dlvsym(), if that is because dlvsym isn't present on apple then I can revise the patch to simplify the logic.


test/meson.build
```diff
@@ -1,3 +1,4 @@
+has_dlvsym = cc.has_function('dlvsym', prefix : '#include<dlfcn.h>')
```
ebassi: I think you need to also define `_GNU_SOURCE`, according to the man page for `dlvsym`:

```
has_dlvsym = cc.has_function(
  'dlvsym',
  prefix: '''#define _GNU_SOURCE
#include <dlfcn.h>''',
)
```

mgorny:
```
commit 756dcaf6a5e07ead098457efba15f5be75bf2123
Author: Eric Anholt <eric@anholt.net>
Date:   Sat Mar 29 00:38:19 2014

    Disable the dlwrap-based tests on apple.
    
    We'd need a dlvsym() equivalent.
```

So yes, please simplify the logic ;-).

By the way, the same file also uses unportable `asprintf()`, so you may want to check for both.

mgorny:
Finally, the whole purpose of that file is to find `dlsym()` with glibc-specific version, so you could just disable it completely on any non-glibc system because it will fail to find it anyway.

rossburton:
Pushed revised commits.

Passing GNU_SOURCE isn't required as has_function doesn't care about headers, but I do need to find libdl as a dependency first.

Removed all traces of build_apple.

ebassi:
Thanks, @rossburton. One last hurdle: as much as I'd like to ignore the Autotools build, we're still supporting it, so `BUILD_APPLE` needs to go away from `configure.ac` and `tests/Makefile.am` as well.

rossburton:
Another patch pushed.  You owe me 🍺 for making me wrestle autoconf after the glory of meson. 😉

PR #161, opened 2018-03-08
test: Avoid egl_without_glx crash on NVIDIA

heftig:
The NVIDIA driver calls dlopen with NULL as filename; strcmp is not
NULL-safe.


nwnk:
Took me a minute to understand the test, but this looks correct. Merged, thanks!

PR #166, opened 2018-04-12
dispatch: assert, not errx, when we have no context

nwnk:
errx() calls exit(), you'd rather get a backtrace.

Signed-off-by: Adam Jackson <ajax@redhat.com>


hadess:
Looks good to me!

Issue #164, opened 2018-03-29
Epoxy fails to detect the (non-epoxy) GLX context under Wayland.


gogglesguy:
This issue has been introduced as part of the 1.5.0 release and as far as I can see this was broken with the following commit: 7c4817f2eed2faf0353c1ceb5f38b500f6aff7cf

Problem: Epoxy fails to detect the current GLX context created by a non-epoxy code.

Work environment is as follows:

1. Running GNOME under wayland ( loginctl showsession c2 -p Type => Type=wayland )
2. Running traditional X11 app using FOX Toolkit (www.fox-toolkit.org) library.
3. FOX does **not** use epoxy, the application **does** use epoxy.
4. The GLX context is created and made current by the FOX library before the application makes any epoxy based calls.

Cause:
1. GLX context set up and made current by non-epoxy library code.
2. application makes a call to _epoxy_has_gl_extension()_
3. As part of _epoxy_internal_has_gl_extension()_, it enters _epoxy_get_bootstrap_proc_address()_ (for glGetString) and  opens the **api.egl_handle** succesfully. The call to _epoxy_egl_get_current_gl_context_api()_ returns EGL_None since there's no EGL Context set up. It falls back to GLX and **api.gl_handle** gets loaded. However **api.glx_handle** is not set at this point.
4. The call to _glGetStringi_ in _epoxy_internal_has_gl_extension()_ will execute _epoxy_current_context_is_glx()_. At this point: **api.egl_handle** is valid and **api.glx_handle** is **null** which results in the function returning false (epoxy_current_context_is_glx:580)

Probably related to #163

gogglesguy:
> I'm thinking the bootstrap code should maybe use dl_iterate_phdr to gently detect whether GLX or EGL is already loaded.

An dlopen() with RTLD_NOLOAD would do the same?

nwnk:
Can you try this branch?

https://github.com/nwnk/libepoxy/tree/even-more-gentle-glx-detection

gogglesguy:
@nwnk yes those changes made it work. 

Before, with the most recent libepoxy 1.5.1 I was getting:
gogglesmm: ../libepoxy/src/dispatch_common.c:862: epoxy_get_proc_address: Assertion `0 && "Couldn't find current GLX or EGL context.\n"' failed.

With your changes, the program runs fine and no longer exits.

ebassi:
PR #168 was merged.

nwnk:
[...]

Issue #160, opened 2018-03-07
1.5.0 breaks KWin


jonnyrobbie:
Updating libepoxy from 1.4.3 to 1.5.0 breaks KWin compositing. OGL 2 nor OGL 3.1 does work. More info can be found here: https://bugs.kde.org/show_bug.cgi?id=391486 and https://bbs.archlinux.org/viewtopic.php?id=235021.

Is this a regression or intentional change? It has been suggested to report upstream, which is here.

ebassi:
Thanks for your report. Unfortunately, I don’t have an nvidia GPU using the binary blob driver, nor I use kwin, so I’ll need your help to investigate it.

It would be great if you could bisect epoxy between the 1.4.3 and 1.5.0 tags, to identify the regression. Reading the issues you linked seems to point to glvnd, but it would be good to be sure.

Additionally, it would be good to understand what the regression actually is, i.e. why is the nvidia driver breaking when testing for the glvnd interface first.

nwnk:
I've added a testing request to the bug:

https://bugs.kde.org/show_bug.cgi?id=391486#c11

ebassi:
@nwnk should we add epoxyinfo to epoxy proper, as a debugging tool?

jonnyrobbie:
Ok, interesting, I tried doing git bisect. The catch is that 1.5.0 did not even build. Part of Arch's PKGBUILD is a check() routine `meson check`

1.5.0 failed

```
ninja: Entering directory `/var/cache/AUR/build/libepoxy/src/build'
ninja: no work to do.
 1/18 header_guards                           OK       0.01 s
 2/18 misc_defines                            OK       0.01 s
 3/18 khronos_typedefs                        OK       0.01 s
 4/18 egl_has_extension_nocontext             OK       0.02 s
 5/18 egl_gl                                  SKIP     0.04 s
 6/18 egl_gles1_without_glx                   SKIP     0.02 s
 7/18 egl_gles2_without_glx                   SKIP     0.01 s
 8/18 glx_beginend                            OK       0.20 s
 9/18 glx_public_api                          FAIL     0.22 s
10/18 glx_public_api_core                     FAIL     0.25 s
11/18 glx_glxgetprocaddress_nocontext         OK       0.10 s
12/18 glx_has_extension_nocontext             OK       0.09 s
13/18 glx_shared_znow                         FAIL     0.25 s
14/18 glx_alias_prefer_same_name              SKIP     0.27 s
[...]

PR #168, opened 2018-04-30
Even more gentle glx detection

nwnk:
This tries even harder to avoid dlopen, essentially only doing so when we would otherwise die for lack of a `GetProcAddress` function or a GL{,ES} implementation. The implementation assumes RTLD_NOLOAD, which is not POSIX, but seems to be available pretty much everywhere these days (cygwin, bionic, solaris, netbsd...).

Fixes issue #164, and I suspect it'll fix issue #160 too.


ebassi:
Looks good to me. If we want to be extra careful with `RTLD_NOLOAD` we can add a test later on.

PR #170, opened 2018-05-01
Make epoxy_load_{egl,glx} load libraries if they exist

nwnk:
This explicitly separates the dlopen and dlsym phases of every library, exposes the GLX and EGL loading logic to the rest of the library, and treats RTLD_NOLOAD and fatal-lookup-error as separate flags.

### DIFF
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -292,7 +292,7 @@ library_init(void)
 }
 
 static bool
-get_dlopen_handle(void **handle, const char *lib_name, bool exit_on_fail)
+get_dlopen_handle(void **handle, const char *lib_name, bool exit_on_fail, bool load)
 {
     if (*handle)
         return true;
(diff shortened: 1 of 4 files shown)


ilovezfs:
@ebassi @nwnk this PR broke libepoxy on macOS. The problem is that `OPENGL_LIB` gets redefined again lower down

https://github.com/nwnk/libepoxy/blob/90db0069b43f40c071510fa4e3b67f6affebd8e2/src/dispatch_common.c#L199

So at run time you get
```
Couldn't open libOpenGL.so.0: dlopen(libOpenGL.so.0, 5): image not found
```

ebassi:
Please, @ilovezfs: open a new issue about this.

ilovezfs:
Where is best? Will do.

ebassi:
@ilovezfs Create [a new issue](https://github.com/anholt/libepoxy/issues/new) on this repo.

ilovezfs:
@ebassi done: https://github.com/anholt/libepoxy/issues/176

PR #179, opened 2018-05-30
Add a CoreGraphics OpenGL test

ebassi:
Building is not enough: we should also run a test to verify that we're not regressing on macOS.


ebassi:
This is based off of #178, so that PR needs to be merged first.

ebassi:
The second test is not working — need to set up Travis to get the test logs on failure. The first test builds and works.

PR #193, opened 2018-10-04
Ensure Autotools enables EGL-X11 support

### LINKED ISSUE #188: Autotools build doesn't actually enable X11
https://github.com/anholt/libepoxy/commit/ce8cbdbe064f5fd7f3e78b6349fa86604e6189d7 made X11 optional for both the Autotools and Meson builds.

[In `src/dispatch_common.h`, we can see](https://github.com/anholt/libepoxy/blob/master/src/dispatch_common.h#L49-L55):

```c
# if !ENABLE_X11
/* Mesa uses this symbol to avoid including X11 headers when including
 * EGL.h; since X11 was explicitly disabled at configuration time, we
 * should do the same
 */
#  define MESA_EGL_NO_X11_HEADERS 1
# endif
```

Unfortunately, the Autotools build never sets `ENABLE_X11`, even when `--enable-x11` was passed, and the dependencies are installed.

So the resulting build is missing some X11 support.

This causes an ABI break when moving from an Autotools build to a Meson one:

```
Functions changes summary: 0 Removed, 0 Changed, 0 Added function
Variables changes summary: 0 Removed, 2 Changed (5 filtered out), 0 Added variables

2 Changed variables:

  [C]'typedef EGLDisplay (typedef EGLNativeDisplayType)* epoxy_eglGetDisplay' was changed at egl_generated.h:991:1:
    type of variable changed:
     in pointed to type 'function type typedef EGLDisplay (typedef EGLNativeDisplayType)':
       parameter 1 of type 'typedef EGLNativeDisplayType' changed:
         underlying type 'void*' changed:
           in pointed to type 'void' at Xlib.h:487:1:
             entity changed from 'void' to 'typedef Display' at Xlib.h:487:1
             type size hasn't changed

  [C]'typedef EGLBoolean (typedef EGLDisplay, EGLNativeDisplayType*)* epoxy_eglQueryNativeDisplayNV' was changed at egl_generated.h:1063:1:
    type of variable changed:
     in pointed to type 'function type typedef EGLBoolean (typedef EGLDisplay, EGLNativeDisplayType*)':
       parameter 2 of type 'EGLNativeDisplayType*' changed:
         pointed to type 'typedef EGLNativeDisplayType' changed at eglplatform.h:117:1, as reported earlier
```

This is when compiling 1.5.2, but I think master is affected in the same way.

Big thanks to @cfergeau for putting me on the right track.

ebassi:
The X11 support in the Autotools build is missing the definition of the
ENABLE_X11 pre-processor symbol. Without it, we're missing the X11
platform definitions for EGL.

Fixes #188

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -157,6 +157,8 @@ if test x$enable_x11 = xno; then
         AC_MSG_ERROR([GLX support is explicitly enabled, but X11 was disabled])
     fi
     build_glx=no
+else
+    AC_DEFINE([ENABLE_X11], [1], [Whether X11 support is enabled])
 fi
 
 AM_CONDITIONAL(BUILD_EGL, test x$build_egl = xyes)


bochecha:
Confirming this fixes #188.

PR #195, opened 2018-10-30
test: Add unit tests for epoxy_gl_version() and epoxy_glsl_version().

anholt:
Based on a bug report I got on IRC.  I thought parsing might be broken on their strings, but it worked fine.  Seems like a good test to have for future reports, though.

### DIFF
--- a/test/gl_version.c
+++ b/test/gl_version.c
@@ -1,0 +1,89 @@
+/*
+ * Copyright © 2018 Broadcom
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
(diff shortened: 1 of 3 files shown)


test/gl_version.c
```diff
@@ -0,0 +1,92 @@
...
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
+ * IN THE SOFTWARE.
+ */
+
+#include <stdio.h>
+#include <assert.h>
+#include "epoxy/gl.h"
+#include "epoxy/glx.h"
```
ebassi: I don't think we need this, and the `glx_common.h` header inclusion; if we do, though, then we need to put this test under the GLX conditional.

test/meson.build
```diff
@@ -33,6 +33,12 @@ test('khronos_typedefs',
                 c_args: common_cflags,
                 dependencies: libepoxy_dep,
                 include_directories: libepoxy_inc))
+test('gl_version',
```
ebassi: We'll also need a change for `Makefile.am`, at least until we drop Autotools support.

PR #203, opened 2019-04-19
enum names

### LINKED ISSUE #202: foo_provider_resolver could be optimized
For extensions, we generate:

```
        case GL_extension_GL_3DFX_tbuffer:
            if (epoxy_conservative_has_gl_extension("GL_3DFX_tbuffer"))
                return epoxy_get_proc_address(entrypoint_strings + entrypoints[i]);
            break;
```

But we already have a copy of the string `GL_3DFX_tbuffer`, in `enum_string`. Or we would, if we didn't format that as `"GL extension \"GL_3DFX_tbuffer\"\0"`. The only reason we format it that way is so we can print that on resolution failure, which isn't especially useful, since a) resolution failure means a broken app anyway and b) the extension type (GL, GLX, EGL) is right there in the name.

So: make the substrings in `enum_string` be just the extension name, and fix the argument to `epoxy_conservative_has_gl_extension` be `enum_string[enum_string_offsets[i]]`. This should both make .rodata smaller since we don't keep two copies of the extension name strings, and shave off a few hundred relocations to the anonymous extension name strings.

anholt:
I saw ajax's suggestion, and it sounded like fun, so here goes.  Tested only with the testsuite, though the testsuite did catch bugs as I built this.  Release build sizes as we go:

```
1121212	  56800	    160	1178172	 11fa3c	master.so
1114236	  56800	    160	1171196	 11defc	providernames.so
1099948	  56800	    160	1156908	 11a72c	enumstrings.so
```

I think the next step would be to sort the providers into "GL versions" and "Extensions" and have the provider resolve loop be "check GL version for providers < cutoff and check GL/GLX/whatever extension for providers >= cutoff.

Fixes #202

### DIFF
--- a/src/gen_dispatch.py
+++ b/src/gen_dispatch.py
@@ -50,6 +50,7 @@ class GLProvider(object):
         self.enum = self.enum.replace(' ', '_')
         self.enum = self.enum.replace('\\"', '')
         self.enum = self.enum.replace('.', '_')
+        self.enum = "PROVIDER_" + self.enum
 
 class GLFunction(object):
     def __init__(self, ret_type, name):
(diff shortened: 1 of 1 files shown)


nwnk:
I like the size reduction, but this doesn't seem to change the number of relocations emitted. Still, lgtm.

PR #208, opened 2019-07-08
Add support for LD_PRELOAD/aliasing.

dwrobel:
On platforms where the LD_PRELOAD is used to overwrite functions
it is needed to look first at the existing (preloaded/aliased) symbol
instead of the real one from the shared library.

For example on Broadcom Nexus platform the following libraries are preloaded:

    - libwayland-egl.so.1
    - libGLESv2.so

where the libwayland-egl.so.1 overwrites eglGetDisplay() from the libGLESv2.so.

In that situation it is inapropriate for libepoxy to select the eglGetDisplay()
from the libGLESv2.so shared library.

Signed-off-by: Damian Wrobel <dwrobel@ertelnet.rybnik.pl>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -49,6 +49,8 @@ AC_LIBTOOL_WIN32_DLL
 AC_PROG_LIBTOOL
 AC_SYS_LARGEFILE
 
+AC_GNU_SOURCE
+
 AC_CHECK_HEADER([KHR/khrplatform.h],
                 [AC_DEFINE([HAVE_KHRPLATFORM_H], [1],
                            [Define to 1 if you have <KHR/khrplatform.h> (used for tests)]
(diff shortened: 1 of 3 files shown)


ebassi:
This change is breaking the build on CI, so I guess it needs further work.

dwrobel:
> This change is breaking the build on CI, so I guess it needs further work.

I saw it it's the `glx_alias_prefer_same_name` test which fails - but it wasn't clear to me what is the purpose of this test. Can it be either a false positive or a broken test?

anholt:
libGLESv2 is exposing egl functions?  That sounds like nexus is really wrong.

Balajims88:
@dwrobel @anholt @ebassi  any update is there?

ebassi:
No updates, and no answer to @anholt's question of November 2019.

PR #210, opened 2019-09-09
test: #define __EXTENSIONS__ on Solaris for strdup()

alanc:
Needed to allow functions outside the original XPG3 standard to be
visible in the Solaris headers when _XOPEN_SOURCE is defined and
not set to a particular value.

Signed-off-by: Alan Coopersmith <alan.coopersmith@oracle.com>


ebassi:
Looks good to me, thanks!

PR #212, opened 2019-11-05
Remove the autotools build system.

anholt:
Meson has been out for a long time, is faster, and is simpler.


gurchetansingh:
Can you make the Meson version 0.46 ;-)?  Will make switching for CrOS easier.

ebassi:
> Can you make the Meson version 0.46 ;-)? Will make switching for CrOS easier.

Is it **really** necessary? Meson 0.46 was released in April 2018. We're already using a long-since deprecated feature (the python3 module) which I was looking forward to removing when we switch to a newer version of Meson and removed Python2 support…

I'm not against this, to be clear, but even Debian stable is running with Meson 0.49.2 as its baseline.

ebassi:
Well, if we want to drop Python 2 by the end of 2019, we can actually do that right now. Then we can also drop Autotools on top of it, and lower the requirement to Meson 0.46, as we don't need to use deprecated modules to find the Python installation.

I've opened #213 for the Python 2 removal.

gurchetansingh:
👍👍 Looks good to me.

Reviewed-by: Gurchetan Singh <gurchetansingh@chromium.org>

PR #213, opened 2019-11-14
Remove Python 2 support

ebassi:
Python 2 is going to reach EOL in January 2020, and most platforms have
already moved to Python 3.


gurchetansingh:
Reviewed-by: Gurchetan Singh <gurchetansingh@chromium.org>

Issue #218, opened 2020-01-29
Regression on Windows


nacho:
I get the following error when building 1.5.4 which I did not get with the 1.5.3:

```
FAILED: test/9f86d08@@gl_version@exe/gl_version.c.obj 
cl @test/9f86d08@@gl_version@exe/gl_version.c.obj.rsp
../libepoxy/test/gl_version.c(79): error C2440: '=': cannot convert from 'const GLubyte *(__cdecl *)(GLenum)' to 'const GLubyte *(__stdcall *)(GLenum)'

```

nacho:
Any chance to get a new release?

ebassi:
Will do a new release by the end of this week (November 27th).

nacho:
Thanks!

PR #220, opened 2020-02-25
Add call convention to mock function

crziter:
A call convention is required to be able to compile tests on Windows MSVC, otherwise the compiler complain about different function signature.


crziter:
This PR fix issue #218

ebassi:
Thanks for fixing this!

PR #225, opened 2020-07-21
Use the darwin_versions argument

ebassi:


chergert:
This seems to fix the linking issues I was having with the new GDK 4 macOS backend.

PR #226, opened 2020-07-21
update EGL registry

eyelash:


danger89:
I also do not understand why there need to be a dedicated branch. But your changes themselves seems fine to me.

PR #228, opened 2020-07-21
GitHub CI

eyelash:
I created a GitHub CI config for libepoxy in case you are interested in using it. The whole build took just over 2 minutes which is probably much faster than Travis. It is also using the Ubuntu 18.04 that is installed on the runner instead of Debian Stretch inside Docker.


ebassi:
Thanks! I was planning to look into this, and you saved me a lot of time. :smiley:

PR #229, opened 2020-07-29
Fix some bugs in loading OpenGL/GLX/EGL libraries

ya-isakov:
Without additional check, even if libOpenGL was loaded, libGL.so will be loaded as well, and used both in gl_handle and glx_handle, so libglvnd libraries will not be used.

The problem is that on Wayland-only system, there will be no libGL, so epoxy_load_gl will fail even if libOpenGL was loaded.

### DIFF
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -189,7 +189,7 @@
 #define GLES2_LIB "libGLESv2.dll"
 #define OPENGL_LIB "OPENGL32"
 #else
-#define GLVND_GLX_LIB "libGLX.so.1"
+#define GLVND_GLX_LIB "libGLX.so.0"
 #define GLX_LIB "libGL.so.1"
 #define EGL_LIB "libEGL.so.1"
 #define GLES1_LIB "libGLESv1_CM.so.1"
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -684,7 +684,8 @@ epoxy_load_gl(void)
     if (!api.gl_handle) {
 	get_dlopen_handle(&api.gl_handle, GLX_LIB, true, true);
 #if PLATFORM_HAS_GLX
-	api.glx_handle = api.gl_handle;
+        if (!api.glx_handle)
+	    api.glx_handle = api.gl_handle;
 #endif
     }
 #endif
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -580,21 +580,28 @@ epoxy_current_context_is_glx(void)
 #else
     void *sym;
 
-    sym = epoxy_conservative_glx_dlsym("glXGetCurrentContext", false);
-    if (sym) {
-        if (glXGetCurrentContext())
-            return true;
-    } else {
-        (void)dlerror();
+    // There is no point of getting symbols from library, which is not loaded,
+    // as handle will always be NULL, when exit_on_fail is false
+    if (api.glx_handle) {
+        sym = epoxy_conservative_glx_dlsym("glXGetCurrentContext", false);
+        if (sym) {
+            if (glXGetCurrentContext())
+                return true;
+        } else {
+            (void)dlerror();
+        }
     }
 
 #if PLATFORM_HAS_EGL
-    sym = epoxy_conservative_egl_dlsym("eglGetCurrentContext", false);
-    if (sym) {
-        if (epoxy_egl_get_current_gl_context_api() != EGL_NONE)
-            return false;
-    } else {
-        (void)dlerror();
+    // Same as for GLX
+    if (api.egl_handle) {
+        sym = epoxy_conservative_egl_dlsym("eglGetCurrentContext", false);
+        if (sym) {
+            if (epoxy_egl_get_current_gl_context_api() != EGL_NONE)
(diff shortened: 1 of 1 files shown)


ya-isakov:
Hmm, it looks like this leads to not loading GLX library, as epoxy_conservative_glx_dlsym is using `exit_if_fails` as a `load` flag.

ebassi:
I guess this is for @nwnk to review.

ya-isakov:
I've found a way to fix problem with loading libGL/libGLX on every `epoxy_current_context_is_glx` check - there is no point of trying `epoxy_conservative_egl_dlsym` if appropriate handle is NULL.
If this handle is NULL, it means that library was not loaded, and `epoxy_conservative_egl_dlsym` with load=exit_if_fails=false will not load it as well, and this will lead to `do_dlsym` also returning NULL. So, I've added checks for api.*_handle in `epoxy_current_context_is_glx` function.

ya-isakov:
Here is output of strace for https://github.com/ebassi/glarea-example on libglvnd-enabled wayland (libGL.so/libGLX.so still exists):
Before my fixes:
....
[pid 3520958] openat(AT_FDCWD, "/usr/lib64/libOpenGL.so.0", O_RDONLY|O_CLOEXEC) = 22
[pid 3520958] openat(AT_FDCWD, "/usr/lib64/libGL.so.1", O_RDONLY|O_CLOEXEC) = 22
[pid 3520958] openat(AT_FDCWD, "/usr/lib64/libGLX.so.0", O_RDONLY|O_CLOEXEC) = 22

After:
...
[pid 3519200] openat(AT_FDCWD, "/usr/lib64/libOpenGL.so.0", O_RDONLY|O_CLOEXEC) = 22

nwnk:
One thing we might want to consider is using `dlopen(RTLD_NOLOAD)` to probe for things...

nwnk:
Patch 1 2 and 4 in this series are:

`Reviewed-by: Adam Jackson <ajax@redhat.com>`

Patch 3 ("Do not load symbols in epoxy_current_context_is_glx [...]") isn't great. With that change, `epoxy_current_context_is_glx` will fail if it's the very first epoxy library call, even if there _is_ a current context and it _is_ GLX. We should instead gently (RTLD_NOLOAD-style) test whether GLX libraries are already loaded, and fill in `api.glx_handle` and friends based on that.

ya-isakov:
@nwnk I'm completely fine, if you would merge only 3 commits. Third one was my best try on fixing the load of libGL.so on every symbol lookup, if EGL is used, but I'm not sure how to fix it properly.

ya-isakov:
Actually, if epoxy_current_context_is_glx if first epoxy library call, it should not fail, just return false. Oh, and without my commit, it will also return false, as it will use `epoxy_conservative_glx_dlsym("glXGetCurrentContext", false)` -> `epoxy_load_glx(false, false)`, and epoxy_load_glx with load=false is not setting api.glx_handle (if library was not loaded before), so it's result will be false, so `epoxy_conservative_glx_dlsym` will return NULL.

TL;DR - without my commit 3, if `epoxy_current_context_is_glx` is the first function to run, it will return false, the same as with my commit.

ya-isakov:
Oh, and how it could be possible, to have GLX context without loading libGLX first? And to load it, application will use `glXCreateNewContext`, which will use `epoxy_glx_dlsym`, which will load libGLX.so, if needed.

nwnk:
> And to load it, application will use glXCreateNewContext, a wrapper for epoxy_glx_dlsym, which will load libGLX.so, if needed, and set api.glx_handle.

This isn't necessarily true. Imagine plugin library A gets loaded first, which links against libGL directly, creates a GLX context, and makes it current. Then plugin library B is loaded, which links against epoxy.

tomeuv:
@nwnk Could we get 1, 2 and 4 merged? With them, I'm able to run virglrenderer without linking to GLX.

ebassi:
Patch 2 introduces a test failure in `glx_alias_prefer_same_name`.

I've opened #238 with patches 1 and 4.

batesj:
Any thoughts on whether https://github.com/anholt/libepoxy/issues/240 is a libepoxy bug or apitrace?

ya-isakov:
@batesj I would say, that it is probably a apitrace not supporting libOpenGL.so. I cannot find any mentions of it in it's code, so, maybe it's trying to override libGL.so, but with my patch, it should not be used, if libOpenGL.so is present in the system

ya-isakov:
@ebassi I'm fine if commit 2 (libGLX.so version fix) is omitted, it's probably different in different distributions. On GLX-free system there is no such file at all...
@nwnk Sorry, proper fixing of patch 3 is way above my skill set in C :(

batesj:
Can there be a new release to pick up this bug fix?

ebassi:
@batesj I just released 1.5.6 which contains the first and last commits of this MR.

ebassi:
If nobody beats me to it, I'll try to carve some time to rework the third commit to follow the approach outlined by @nwnk.

batesj:
Any thoughts on bringing these fixes back without breaking some distros? Compile option perhaps?

PR #231, opened 2020-10-14
meson: Set same variables in declare_dependency() as in pkgconfig

xclaesse:
GTK currently fails to detect if epoxy has been built with EGL on
Windows when epoxy is a subproject. To fix that it needs to get that
information from the dependency variables.

### DIFF
--- a/src/meson.build
+++ b/src/meson.build
@@ -72,16 +72,21 @@ libepoxy = library(
   link_args: common_ldflags,
 )
 
+epoxy_has_glx = build_glx ? '1' : '0'
+epoxy_has_egl = build_egl ? '1' : '0'
+epoxy_has_wgl = build_wgl ? '1' : '0'
+
 libepoxy_dep = declare_dependency(
   link_with: libepoxy,
   include_directories: libepoxy_inc,
   dependencies: epoxy_deps,
   sources: epoxy_headers,
+  variables: {
+    'epoxy_has_glx': epoxy_has_glx,
+    'epoxy_has_egl': epoxy_has_egl,
+    'epoxy_has_wgl': epoxy_has_wgl,
+  },
 )
-
-epoxy_has_glx = build_glx ? '1' : '0'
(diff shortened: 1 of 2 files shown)


xclaesse:
See https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/2691

src/meson.build
```diff
@@ -72,17 +72,22 @@ libepoxy = library(
...
+epoxy_has_glx = build_glx ? '1' : '0'
+epoxy_has_egl = build_egl ? '1' : '0'
+epoxy_has_wgl = build_wgl ? '1' : '0'
+
 libepoxy_dep = declare_dependency(
   link_with: libepoxy,
   include_directories: libepoxy_inc,
   dependencies: epoxy_deps,
   sources: epoxy_headers,
+  variables: {
```
ebassi, line 84: The usage of a dictionary might require a bump in the required version of Meson. Right now it's at 0.48, but I'm amenable to changing it to the minimum version that supports this.

xclaesse: Good point, it actually requires 0.54, updated the PR.

I'm actually disappointed that declare_dependency() takes variables as dict and pkg.generate() takes variables as list of "var=val", which means I had to duplicate that information in this PR. Ideally meson should accept both formats so we could have a common variable between boths.

xclaesse:
@ebassi need to update Meson version in CI, do you know how?

ebassi:
@xclaesse I'll have to update the Docker image we use on Travis.

xclaesse:
FYI, I fixed the inconsistency in meson: https://github.com/mesonbuild/meson/pull/7853.

ebassi:
We don't use Travis any more, so no more Docker images. Care to rebase on `master`?

xclaesse:
I miss gitlab's rebase button. Done manually :)

ebassi:
Thanks!

PR #232, opened 2020-10-24
dispatch: fix loading of GL without GLX library

pks-t:
When loading the GL library in `epoxy_load_gl`, we in turn try to load
OPENGL_LIB and GLX_LIB. But notably, we'll also try to load GLX_LIB even
if loading OPENGL_LIB succeeded. As this load will abort in case the
library wasn't found, the end result is that systems without GLX_LIB
will always fail `epoxy_load_gl`.

Fix the issue by only trying to load GLX_LIB in case OPENGL_LIB wasn't
found.

### DIFF
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -675,8 +675,10 @@ epoxy_load_gl(void)
 	get_dlopen_handle(&api.gl_handle, OPENGL_LIB, false, true);
 #endif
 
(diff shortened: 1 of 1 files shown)


ebassi:
See #257 and the discussion in #252.

ebassi:
This should have been fixed in #257.

PR #233, opened 2020-11-05
Fix meson c_std error

c72578:
Fixes the following error, when building with meson 0.56.0:
```console
meson.build:1:0: ERROR: Value "gnu99" (of type "string") for combo
option "C language standard to use" is not one of the choices.
Possible choices are (as string): "none", "c89", "c99", "c11".

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
 project('libepoxy', 'c', version: '1.5.5',
         default_options: [
           'buildtype=debugoptimized',
-          'c_std=gnu99',
+          'c_std=c99',
           'warning_level=1',
         ],
         license: 'MIT',


ebassi:
Raising an error without any deprecation warning prior is not very nice. Prior to Meson 0.56, `gnu99` automatically degraded to `c99`.

Which compiler are you using, just so I know?

xclaesse:
Oops, that's a known regression in Meson that we forgot to fix before the release: https://github.com/mesonbuild/meson/issues/7611. That will definitely have to be fixed for 0.56.1.

c72578:
@ebassi The compiler, where this was observed, was MSVC. See further details here at vcpkg:
https://github.com/microsoft/vcpkg/pull/14344#issuecomment-721014658
@xclaesse Thanks for keeping this regression on a list to be fixed in meson 0.56.1

c72578:
A PR to fix this in meson is currently under discussion / in preparation:
https://github.com/mesonbuild/meson/pull/7862
So I will close the PR here.

c72578:
Upstream fixes are now in the 0.56 branch:
https://github.com/mesonbuild/meson/commit/fd6585ecc41e1c731224cac08a3f6ef6f49ab184
and in master:
https://github.com/mesonbuild/meson/commit/333e0aeef4b2cc7be82a89c3a1a5a4a4b9af2b53

PR #239, opened 2021-02-11
Support ANGLE on macOS

akihikodaki:


lbibass:
Any updates on getting this pulled in?

meson.build
```diff
@@ -174,12 +177,12 @@ x11_dep = dependency('x11', required: false)
 # if we fail
 gles2_dep = dependency('glesv2', required: false)
 if not gles2_dep.found()
-  gles2_dep = cc.find_library('libGLESv2', required: false)
+  gles2_dep = cc.find_library('GLESv2', required: false)
```
ebassi, line 180: Has this change been tested on systems that do not ship a pkg-config file and are not macOS?

AFAIR, Windows with ANGLE is a possible platform where this might be a problem.

ebassi: Let's ask @fanc999 about this particular check.

fanc999: Yup, it will be a problem on Windows, as that is the .lib file name that libANGLE gives us.

Perhaps we can cycle through a list of known names for the library, or select the lib name using the target OS platform.

akihikodaki: @fanc999 I have just observed the behavior of `cc.find_library` with ANGLE where `is_component_build=true` (i.e. ANGLE is built as shared library, which is the default configuration):
- Microsoft's `link` only accepted `libGLESv2.dll`.
- MinGW64 (from Git for Windows SDK) accepted all of `GLESv2`, `libGLESv2` and `libGLESv2.dll`.

Can you check your configuration and see if it accepts `GLESv2` and `libGLESv2.dll`? I think the best bet is to have `GLESv2` and `libGLESv2.dll` so I want to confirm if they work.

lbibass:
Checking in on getting this merged into master again! Would really love to see it.

ebassi:
This will not get merged, unless the issue on Windows is solved.

meson.build
```diff
@@ -165,6 +165,9 @@ endif
 dl_dep = cc.find_library('dl', required: false)
 gl_dep = dependency('gl', required: false)
 egl_dep = dependency('egl', required: false)
+if not egl_dep.found()
+  egl_dep = cc.find_library('EGL', required: false)
+endif
```
t0rr3sp3dr0, line 170:
```suggestion
if not egl_dep.found()
  egl_dep = cc.find_library('EGL', required: false)
endif
if not egl_dep.found()
  egl_dep = cc.find_library('libEGL', required: false)
endif
```

meson.build
```diff
@@ -174,12 +177,12 @@ x11_dep = dependency('x11', required: false)
 # if we fail
 gles2_dep = dependency('glesv2', required: false)
 if not gles2_dep.found()
-  gles2_dep = cc.find_library('libGLESv2', required: false)
+  gles2_dep = cc.find_library('GLESv2', required: false)
 endif
```
t0rr3sp3dr0, line 181:
```suggestion
if not gles2_dep.found()
  gles2_dep = cc.find_library('GLESv2', required: false)
endif
if not gles2_dep.found()
  gles2_dep = cc.find_library('libGLESv2', required: false)
endif
```

meson.build
```diff
@@ -174,12 +177,12 @@ x11_dep = dependency('x11', required: false)
...
 
 gles1_dep = dependency('glesv1_cm', required: false)
 if not gles1_dep.found()
-  gles1_dep = cc.find_library('libGLESv1_CM', required: false)
+  gles1_dep = cc.find_library('GLESv1_CM', required: false)
 endif
```
t0rr3sp3dr0, line 186:
```suggestion
if not gles1_dep.found()
  gles1_dep = cc.find_library('GLESv1_CM', required: false)
endif
if not gles1_dep.found()
  gles1_dep = cc.find_library('libGLESv1_CM', required: false)
endif
```

t0rr3sp3dr0 (commented):
Would these suggested changes be enough to get this PR merged?

akihikodaki:
> Would these suggested changes be enough to get this PR merged?

I need an answer for https://github.com/anholt/libepoxy/pull/239#discussion_r651877285.

meson.build
fanc999: Oh,

I'm so sorry, somehow this slipped out of my radar...  To be exact, it is `libGLESv2.lib`, so it needs to be `libGLESv2` for the `cc.find_library()` call.  The way to get away from this is to use a Windows 10+ `GLESv2.lib` symlink to `libGLESv2.lib`, or make a copy of `libGLESv2.lib`.

akihikodaki: @fanc999 Actually it's named `libGLESv2.dll.lib` on my side. Did you compile ANGLE with `is_component_build=false`? I didn't think it is supported by ANGLE developers and is compatible with libepoxy, but my past understanding may be wrong.

akihikodaki: I was wrong about `is_component_build`. It was an option for Chromium builds, and standalone ANGLE comes with both of dynamic libraries and static libraries.

I decided to dig a bit deeper and here are the results:
- The upstream ANGLE names dynamic libraries as `lib*.dll.lib` and static libraries as `obj/lib*/lib*_static.lib`. You can see the code naming dynamic libraries at: https://chromium.googlesource.com/chromium/src/build/+/e8e1d878ba228409f23491766deb9a7c8c10deb7%5E%21/#F7
- Qt 5 comes with its own build infrastructure as you can see at: https://code.qt.io/cgit/qt/qtbase.git/tree/src/angle?h=5.15.2. That is the reason why their builds have a different suffix. Fortunately, Qt ships with CMake import libraries so I rewrote to use them. It also allows to pick up header files automatically as a bonus. @fanc999 Does this look good to you?

fanc999: Hi,

Thanks for the work, however I think it is still best to just look through for the library names for EGL and GLESv2, instead of using the CMake files for QT-5.x, since:

*  It may seem confusing for people (Qt5Gui is rather large).
*  People might have opted to build QT-5.x without libANGLE.

My take at this.

With blessings, thank you!

akihikodaki: @fanc999 What about explicitly requiring that Qt 5 provides ANGLE feature instead of just checking Qt 5's existence? I'm trying to avoid just specifying the file name because it barely tells who provides the file and is fragile to environment variances as we see now with this pull request.

fanc999: Hi @akihikodaki,

I understand your point, but this is unfortunately the case with things on Visual Studio as .lib names can vary very often, and it really depends on the dependencies' or CMake/Meson upstream to see whether we have one or more of the following:

*  pkg-config (.pc) files
*  CMake config files
*  Built-in detection scheme of packages in the build system (i.e. in Meson and/or CMake, examples include openssl, libpng and zlib)

IMO, I still believe that pulling in the CMake config files for Qt5Gui is making things a bit too big, due to the numerous options that people might use for it (building QT-5.x with `opengl dynamic` seems to also disable `opengles2` etc, but it does pull in the `libEGL.lib` and `libGLESv2.lib`[1]).

Thus, I think:

*  Look for `GLESv2`/`EGL` for the `find_library()` call on non-Visual Studio-style builds (i.e. `cc.get_argument_syntax() != 'msvc'`)
*  In a small for-loop, look for `libGLESv2` and `libGLESv2.dll`/`libEGL` and `libEGL.dll` for Visual Studio-style builds.  The alternative is to simply ask people in the docs to copy or symlink `libGLESv2.dll.lib` to/from `libGLESv2.lib`, and likewise for the EGL library.

[1]: 
[Qt5Gui-cmake-config.zip](https://github.com/anholt/libepoxy/files/7689873/Qt5Gui-cmake-config.zip), my QT-5.15.2 Qt5GUI CMake config files as reference.

With blessings, thank you!

akihikodaki: @fanc999
> I understand your point, but this is unfortunately the case with things on Visual Studio as .lib names can vary very often, and it really depends on the dependencies' or CMake/Meson upstream to see whether we have one or more of the following:

Filenames are fragile in nature and that is true whether they are for Visual Studio or Unix tools. pkg-config and CMake import libraries intend to be solutions for that occasions and provide more stable identifiers so they should fit well for this use case. We already rely on pkg-config for Mesa and the additional complexity caused by it is not considered problematic.

> IMO, I still believe that pulling in the CMake config files for Qt5Gui is making things a bit too big, due to the numerous options that people might use for it (building QT-5.x with opengl dynamic, which I believe is the default in the official QT binaries, seems to also disable opengles2 etc, but it does pull in the libEGL.lib and libGLESv2.lib[1]).

Qt 5 is indeed big, but we are dealing cases where users intend to use the OpenGL implementation it provides anyway and I don't see the reason we have to be concerned about the details, including the options they give. In the case of `opengl dynamic` with `opengles2` disabled, both of CMake and `cc.find_library()` will find those `libEGL.lib` and `libGLESv2.lib` and use them so there should be no difference. Is this convincing for you, or you have an idea of some problematic scenario?

fanc999: Hi,

If you have a to simple enough means to check that Qt5Gui is indeed built with libANGLE support, and if the maintainers are fine with using Qt5Gui as a means of checking for GLES/EGL, then I think I am fine, then.

Thanks for your patience!

akihikodaki: Explicitly specified EGL and GLESv2 targets with d1e471bf9b70ea7db8599e24cf9bd3b227145a84. Actually this avoids linking with irrelevant Qt 5 libraries and would be a nice improvement.

akihikodaki:
This pull request should be now ready for reviews by maintainers.

danger89:
@ebassi please take a second look.

akihikodaki:
8611b075ec8428d45f54cc58273f5330f842cbb0: rebased to c84bc9459357a40e46e2fec0408d04fbdde2c973.

akihikodaki:
@ebassi Hi, it's been three years since this was submitted. Can you take a look at this?

meson.build
```diff
@@ -165,25 +165,47 @@ endif
...
+# We have multiple checks for EGL and GLES v2/v1 to support different providers:
+# 1. pkg-config for Mesa
+# 2. find_library() for ANGLE, which do not support pkg-config nor CMake.
+#    Note that Microsoft's "link" requires "lib" prefix.
+# 3. CMake for Qt 5, which bundles ANGLE.
+egl_dep = dependency('egl', required: false)
+if not egl_dep.found()
+  egl_dep = cc.find_library('EGL', required: false)
+endif
+if not egl_dep.found()
```
ebassi, line 182:
```suggestion
if not egl_dep.found() and host_machine == 'windows'
```

meson.build
```diff
@@ -165,25 +165,47 @@ endif
...
+  egl_dep = dependency('Qt5Gui', modules: 'Qt5::Gui_EGL', required: false)
+endif
+
 gles2_dep = dependency('glesv2', required: false)
 if not gles2_dep.found()
-  gles2_dep = cc.find_library('libGLESv2', required: false)
+  gles2_dep = cc.find_library('GLESv2', required: false)
+endif
+if not gles2_dep.found()
```
ebassi, line 193:
```suggestion
if not gles2_dep.found() and host_machine == 'windows'
```

meson.build
```diff
@@ -165,25 +165,47 @@ endif
...
+  egl_dep = dependency('Qt5Gui', modules: 'Qt5::Gui_GLESv2', required: false)
 endif
 
 gles1_dep = dependency('glesv1_cm', required: false)
 if not gles1_dep.found()
-  gles1_dep = cc.find_library('libGLESv1_CM', required: false)
+  gles1_dep = cc.find_library('GLESv1_CM', required: false)
 endif
+if not gles1_dep.found()
```
ebassi, line 204:
```suggestion
if not gles1_dep.found() and host_machine == 'windows'
```

ebassi:
Two minor changes to limit DLL checks on Windows.

akihikodaki:
> Two minor changes to limit DLL checks on Windows.

Applied with commit b0d6d609d64dcf71845801233d1d3680b9d50288.

meson.build
```diff
@@ -165,25 +165,47 @@ endif
...
+# We have multiple checks for EGL and GLES v2/v1 to support different providers:
+# 1. pkg-config for Mesa
+# 2. find_library() for ANGLE, which do not support pkg-config nor CMake.
+#    Note that Microsoft's "link" requires "lib" prefix.
+# 3. CMake for Qt 5, which bundles ANGLE.
+egl_dep = dependency('egl', required: false)
+if not egl_dep.found()
+  egl_dep = cc.find_library('EGL', required: false)
+endif
+if not egl_dep.found() and host_machine == 'windows'
```
Fahien, line 182: Testing on darwin with meson `1.4.0`:
```
../meson.build:182:43: ERROR: 
Trying to compare values of different types (MachineInfo, str) using ==.
This was deprecated and undefined behavior previously and is as of 0.60.0 a hard error.
```

I suppose this can be fixed with `host_machine.system() == 'windows'` here, and the other two lines below?

Fahien: Actually, you can use `host_system` already defined above :)

avoidik, line 182:
```suggestion
if not egl_dep.found() and host_system == 'windows'
```

akihikodaki:
9fdedad23ee99031401bb333d2e56be0187f70b1: Replaced `host_machine` with `host_system`.

avoidik:
@ebassi could you please approve this workflow?

AkihiroSuda:
> This will not get merged, unless the issue on Windows is solved.

@ebassi Is this issue solved?

cc @anholt

AkihiroSuda:
What's the current status of this?

src/meson.build
```diff
@@ -55,7 +55,7 @@ endif
 # Maintain compatibility with autotools; see: https://github.com/anholt/libepoxy/issues/108
 darwin_versions = [1, '1.0']
 
-epoxy_deps = [ dl_dep, ]
+epoxy_deps = [ dl_dep, egl_dep ]
```
mohd-akram: This isn't right. It causes EGL to always get added to `Requires.private` in pkgconfig even if it was not requested (but happened to be found). The library is loaded dynamically so it's not a dep.

mohd-akram (changes requested):

src/meson.build
akihikodaki: I reverted the change with commit 47d975d726ece5747333ccd9c96d2d76ac753924.

meson.build
```diff
@@ -168,24 +168,42 @@ gl_dep = dependency('gl', required: false)
 if not gl_dep.found() and not build_glx
   gl_dep = dependency('opengl', required: false)
 endif
+
+# We have multiple checks for EGL and GLES v2/v1 to support different providers:
+# 1. pkg-config for Mesa
+# 2. find_library() for ANGLE, which do not support pkg-config nor CMake.
```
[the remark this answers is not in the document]
akihikodaki: I dropped the `find_library()` with commit 47d975d726ece5747333ccd9c96d2d76ac753924.

ebassi (approved):
Not super happy about having to check for Qt, but that's them breaks, I guess…

PR #243, opened 2021-03-06
gen_dispatch.py: close output files

amezin:
So all buffered data actually gets written.

Fixes https://github.com/anholt/libepoxy/issues/242


EmadAlblueshi:
Thank you @amezin this made the `libepoxy-1.5.5` compilation succeeded!

Hi @ebassi I think this should be merged to fix the issue in Windows

Issue #246, opened 2021-03-21
GLsizeiptr and GLintptr defined incorrectly?


bemoody:
According to the GL specs, the types GLsizeiptr and GLintptr should be the size of a pointer (as the 'ptr' suffix would suggest.)

This is supported by the <GL/glext.h> supplied by the MinGW-w64 cross compiler (mingw-w64-common 6.0.0-3 on Debian):

    typedef ptrdiff_t GLsizeiptr;
    typedef ptrdiff_t GLintptr;
    ...
    typedef ptrdiff_t GLsizeiptrARB;
    typedef ptrdiff_t GLintptrARB;

libepoxy 1.5.3 agrees:

    typedef ptrdiff_t GLintptr;
    typedef ptrdiff_t GLsizeiptr;
    ...
    typedef ptrdiff_t GLintptrARB;
    typedef ptrdiff_t GLsizeiptrARB;

However, libepoxy 1.5.5 disagrees:

    typedef long khronos_intptr_t;
    typedef long khronos_ssize_t;
    ...
    typedef khronos_intptr_t GLintptr;
    typedef khronos_intptr_t GLintptrARB;
    typedef khronos_ssize_t GLsizeiptr;
    typedef khronos_ssize_t GLsizeiptrARB;

Which is correct?  What does the implementation of, say, glBindBuffersRange() expect, on 64-bit Windows?

ebassi:
The header is generated from `gl.xml`, which says:

```xml
<type requires="khrplatform">typedef khronos_intptr_t <name>GLintptr</name>;</type>
<type requires="khrplatform">typedef khronos_ssize_t <name>GLsizeiptr</name>;</type>
```

but, sadly, not every system provides a `khrplatform.h`, and we can't detect if an header exists before including it; this means we get to define a bunch of types that should be in that header.

I guess it should be possible to change those definitions depending on the platform.

ebassi:
Could you try #247 and see if it fixes the problem? I don't have a Windows 64 bit build I can test on, at the moment.

bemoody:
Thanks for the quick response!  At the moment I don't have a Windows
system handy either.  Actually trying to come up with a simple and
definitive test case is a little tricky as well... for one thing, it
seems that Mesa/Gallium has problems if you try to create a single
buffer larger than 4 GiB.

One simple test that does work with Mesa is to try running:

    GLint buffer;
    glGenBuffers(1, &buffer);
[...]

PR #247, opened 2021-03-21
Support Win64 pointer-sized types

ebassi:
Since we're generating stub types from khrplatform.h because we can't
include it reliably, we'll have to deal with some of the fallout of the
Win64 types.

Fixes: #246


ebassi:
Since this change is merely reproducing things that are already in the official `khrplatform.h` header, I think it's safe to merge.

PR #248, opened 2021-04-19
build: Check for included system headers

ebassi:
The generated dispatch headers for EGL and GLX include system headers.
We do a pkg-config check for those, but the dependency is not required,
in order to support builds on systems that do not use pkg-config.

If the dependency is not found, we need to check for the headers we
include, otherwise building Epoxy will fail. If the support for EGL or
GLX was explicitly enabled, fail the configuration step; otherwise,
disable the EGL or GLX support.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -69,6 +69,63 @@ else
   has_znow = true
 endif
 
+# Dependencies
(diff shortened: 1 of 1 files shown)


ebassi:
Should be fixed by #261

PR #255, opened 2021-06-16
gen_dispatch: Define KHR platform types only if header not present

kkonopko:
On some platforms, `khrplatform.h` is present and provides information
(basic types, pre-processor definitions) that make it subtly different
than what's auto-generated by the libepoxy generator script
(`gen_dispatch.py`).  This later results in compilation failures for
such platforms.

Since it's possible to determine whether the header is present, use the
logic to include it preferably, and fall back to the original behaviour
otherwise.

### DIFF
--- a/src/gen_dispatch.py
+++ b/src/gen_dispatch.py
@@ -501,11 +501,12 @@ class Generator(object):
                 self.outln('#endif')
                 self.outln('#endif')
         else:
+            self.outln('#ifdef HAVE_KHRPLATFORM_H')
+            self.outln('#include <KHR/khrplatform.h>')
+            self.outln('#else')
             # Add some ridiculous inttypes.h redefinitions that are
-            # from khrplatform.h and not included in the XML.  We
-            # don't directly include khrplatform.h because it's not
-            # present on many systems, and coming up with #ifdefs to
-            # decide when it's not present would be hard.
+            # from khrplatform.h and not included in the XML when
+            # it's not present on the platform.
             self.outln('#define __khrplatform_h_ 1')
             self.outln('typedef int8_t khronos_int8_t;')
             self.outln('typedef int16_t khronos_int16_t;')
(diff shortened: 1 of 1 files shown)


src/gen_dispatch.py
```diff
@@ -501,11 +501,12 @@ def write_header(self, out_file):
                 self.outln('#endif')
                 self.outln('#endif')
         else:
+            self.outln('#ifdef HAVE_KHRPLATFORM_H')
```
ebassi, line 504: This relies on everyone using Epoxy to:

 1. check if `khrplatform.h` is available
 2. define `HAVE_KHRPLATFORM_H` as part of their build, either by:
   - injecting something like `-DHAVE_KHRPLATFORM_H` in their compiler flags *or*
   - only ever including Epoxy headers after they include a "config.h" which defines that symbol

Which is somewhat suboptimal for everything that is not a leaf application. For instance, GTK exposes `epoxy/gl.h` in its public headers, which means everyone using GTK would now have to either use the existing fallback (most likely) or do the `khrplatform.h` discovery, because people will build Epoxy on one system, and build code against Epoxy on another.

That's why the comment says: "coming up with #ifdefs to decide when it's not present on the platform would be hard". It's not hard to add a pre-processor symbol: it's hard to do so while without making every downstream use of Epoxy more complicated.

kkonopko: Yes, this patch _does_ rely on `HAVE_KHRPLATFORM_H` ending up in `config.h`.

> [...] because people will build Epoxy on one system, and build code against Epoxy on another.

Oh, that's a bit of a surprise to me.  Maybe I'm missing something but doing this may introduce subtle inconsistencies, no?  If the platform on which libepoxy is built have some subtle differences in EGL APIs (types, pre-procesor macros and whatnot) to those on the platform where it's actually used (and linked against), doesn't it result in any problems?

danger89: Any update on this progress?

PR #256, opened 2021-06-16
Add EGL include directories

kkonopko:
Although libepoxy includes EGL as its (optional) dependency, it does not
respect the include directory flags required by that dependency.  On
some platforms like Broadcom, this path is non-standard and libepoxy
needs to take that into account.

This patch ensures that libepoxy includes any flags for the platform
EGL.


ebassi:
Fixed by #261

Issue #253, opened 2021-05-21
glGetString displays (null) for nvidia-340xx


wojnilowicz:
Steps to reproduce:
1) unpack [gl3.zip](https://github.com/anholt/libepoxy/files/6522638/gl3.zip)
2) compile gl3.cxx with "cc -o gl3 gl3.cxx -Dwith_epoxy=1 -lGL -lX11 -lepoxy"
3) run "./gl3"

Result:
OpenGL vendor string: (null)

Expected result:
OpenGL vendor string: NVIDIA Corporation

Additional info:
1) compiling gl3.cxx with "cc -o gl3 gl3.cxx -Dwith_epoxy=0 -lGL -lX11 -lepoxy" gives
OpenGL vendor string: NVIDIA Corporation

2) compiling gl3.cxx with "cc -o gl3 gl3.cxx -Dwith_epoxy=1 /usr/lib64/libOpenGL.so /usr/lib64/libGLX.so -lX11 -lepoxy" gives
OpenGL vendor string: (null)

3) compiling gl3.cxx with "cc -o gl3 gl3.cxx -Dwith_epoxy=0 /usr/lib64/libOpenGL.so /usr/lib64/libGLX.so -lX11 -lepoxy" gives
libGL error: No matching fbConfigs or visuals found
libGL error: failed to load driver: swrast
X Error of failed request:  GLXBadFBConfig
  Major opcode of failed request:  154 (GLX)
  Minor opcode of failed request:  0 ()
  Serial number of failed request:  37
  Current serial number in output stream:  37

4) reverting https://github.com/anholt/libepoxy/commit/dbfa4b209c0712b67dfeb8366c2ebe8503063b52 helps

5) KDE relies on info returned from glGetString to enable compositing, and after getting (null) for every query, the compositor remains disabled.

6) Fedora 34 X86_64, libepoxy 1.5.7, nvidia 340.108

Please fix this.

ebassi:
This is a duplicate of #252.

ya-isakov:
@wojnilowicz is it the same with `cc -o gl3 gl3.cxx -Dwith_epoxy=1 -lX11 -lepoxy` (just dropped linking with libOpenGL.so and libGLX.so, libepoxy should load them dynamically)

ya-isakov:
@wojnilowicz also, could you, please, check if libGLX.so is provided by libglvnd?

wojnilowicz:
[...]

Issue #252, opened 2021-05-13
Upgrading libepoxy from 1.5.5 to 1.5.7 results in Xorg crashing on boot


Ernest1338:
Hello,

Basic info:
- OS: Manjaro Linux KDE
- Kernel: 5.11.18-1-MANJARO
- GPU: Radeon R9 380 with AMDGPU drivers

I encountered this issue after an recent Manjaro update which amongst other things updated the libepoxy version from 1.5.5 to 1.5.7 which (as I later found) was the package causing my Xorg to crash on boot.
Using the ```git bisect``` I managed to find the problematic commit, which is: dbfa4b2

This is the Xorg log after it crashed:
```
[  5840.928] (WW) Failed to open protocol names file lib/xorg/protocol.txt
[  5840.929] 
X.Org X Server 1.20.11
... a lot more stuff (I can post if needed)
[  5841.290] (II) Initializing extension XFree86-DRI
[  5841.290] (II) Initializing extension DRI2
[  5841.291] (II) AMDGPU(0): Setting screen physical size to 1377 x 285
[  5841.311] (EE) 
[  5841.311] (EE) Backtrace:
[  5841.311] (EE) 0: /usr/lib/Xorg (xorg_backtrace+0x53) [0x55f225222fd3]
[  5841.311] (EE) 1: /usr/lib/Xorg (0x55f2250dc000+0x151df5) [0x55f22522ddf5]
[  5841.311] (EE) 2: /usr/lib/libc.so.6 (0x7f2164329000+0x3cf80) [0x7f2164365f80]
[  5841.311] (EE) 3: /usr/lib/libc.so.6 (gsignal+0x145) [0x7f2164365ef5]
[  5841.311] (EE) 4: /usr/lib/libc.so.6 (abort+0x116) [0x7f216434f862]
[  5841.311] (EE) 5: /usr/lib/libc.so.6 (0x7f2164329000+0x26747) [0x7f216434f747]
[  5841.311] (EE) 6: /usr/lib/libc.so.6 (0x7f2164329000+0x35646) [0x7f216435e646]
[  5841.311] (EE) 7: /usr/lib/xorg/modules/drivers/amdgpu_drv.so (0x7f216485f000+0x8fd3) [0x7f2164867fd3]
[  5841.311] (EE) 8: /usr/lib/xorg/modules/drivers/amdgpu_drv.so (0x7f216485f000+0x933a) [0x7f216486833a]
[...]

Issue #240, opened 2021-03-02
Draw calls missing in apitrace (regression)


linkmauve:
dbfa4b209c0712b67dfeb8366c2ebe8503063b52 is the first bad commit.

This happens with apitrace master, gtk4-demo records everything (?) except draw calls like `glDrawArrays()`.

batesj:
It looks like before that commit, there was a bug that caused libGL.so to always be used even if libOpenGL.so was present. After the commit, libOpenGL.so will be used if it is found, else libGL.so will be used.

ya-isakov:
@linkmauve could you please check, if there is libOpenGL.so.0, and move it somewhere to test, if it will fix your issue? My guess is that apitrace doesn't support libglvnd.

ya-isakov:
Also it is worthy to check if setting TRACE_LIBGL=libOpenGL.so.0 will help as well

shadeslayer:
Hi
Could you provide more steps to reproduce the issue? I've built libglvnd, libepoxy, mesa ( with the -Dglvnd=true flag) and apitrace locally and can't seem to reproduce the issue with glmark2. I definitely see glDrawArray calls in my traces.

evelikov:
@ya-isakov libglvnd should just work, even if libepoxy uses `libOpenGL.so` while apitrace does `libGL.so`. Albeit small, there's a possibility of this might be a libglvnd bug in itself. 

If it were me, first I would check if apitrace wraps `dlsym` and `{egl,glX}GetProcAddress` to intercept all GL dispatch (glDrawArray or otherwise) and see in which library we end on the problematic call(s).

PR #257, opened 2021-06-17
Allow libopengl.so to be used when GLX_LIB is missing

batesj:
This maintains compatibility with previous behavior of
always using GLX_LIB if it is found. The only change is
when there is no GLX_LIB.

Previous behavior when no GLX_LIB:
- abort.

New behavior when no GLX_LIB:
- Try to load libOpenGL.so as gl_handle (glx_handle remains NULL).
- Else, abort.

### DIFF
--- a/src/dispatch_common.c
+++ b/src/dispatch_common.c
@@ -670,13 +670,23 @@ epoxy_load_gl(void)
     get_dlopen_handle(&api.gl_handle, OPENGL_LIB, true, true);
 #else
 
+    // Prefer GLX_LIB over OPENGL_LIB to maintain existing behavior.
+    // Using the inverse ordering OPENGL_LIB -> GLX_LIB, causes issues such as:
+    // https://github.com/anholt/libepoxy/issues/240 (apitrace missing calls)
+    // https://github.com/anholt/libepoxy/issues/252 (Xorg boot crash)
+    get_dlopen_handle(&api.glx_handle, GLX_LIB, false, true);
+    api.gl_handle = api.glx_handle;
+
 #if defined(OPENGL_LIB)
     if (!api.gl_handle)
-	get_dlopen_handle(&api.gl_handle, OPENGL_LIB, false, true);
+        get_dlopen_handle(&api.gl_handle, OPENGL_LIB, false, true);
 #endif
 
-    get_dlopen_handle(&api.glx_handle, GLX_LIB, true, true);
-    api.gl_handle = api.glx_handle;
+    if (!api.gl_handle) {
+        fprintf(stderr, "Couldn't open %s or %s\n", GLX_LIB, OPENGL_LIB);
+        abort();
+    }
+
 #endif
 }
 


batesj:
@ya-isakov @tomeuv @ebassi 

PTAL. Also, if you have a repro with one of the known downstream failures caused by #229, it would help to have someone test one of those cases.

batesj:
ping.

ya-isakov:
@batesj Checked it on my libGL-free system - it works:
```
openat(AT_FDCWD, "/usr/lib64/libGL.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib64/libOpenGL.so.0", O_RDONLY|O_CLOEXEC) = 21
openat(AT_FDCWD, "/lib64/libGLX.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib64/libGLX.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/lib64/libGL.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib64/libGL.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/lib64/libGLX.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib64/libGLX.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/lib64/libGL.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib64/libGL.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
```
Without your patch, vanilla libepoxy:
```
openat(AT_FDCWD, "/usr/lib64/libGLdispatch.so.0", O_RDONLY|O_CLOEXEC) = 16
openat(AT_FDCWD, "/usr/lib64/libOpenGL.so.0", O_RDONLY|O_CLOEXEC) = 21
....
openat(AT_FDCWD, "/usr/lib64/haswell/libGL.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib64/x86_64/libGL.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/lib64/libGL.so.1", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
write(2, "Couldn't open libGL.so.1: libGL."..., 96Couldn't open libGL.so.1: libGL.so.1: cannot open shared object file: No such file or directory
```
But previous one worked for me, too, so, somebody with Xorg should also check...

batesj:
I've tested this extensively now in Chrome OS and confirmed that it works to enable the use of libOpenGL.so while not changing behavior of existing libGLX expectations. For example, confirmed that this libepoxy change does not break crostini apps when libopengl0 is installed.

batesj:
@ebassi ptal

ebassi:
Let's see if the people having issues in #252 and #240 can test this: @Ernest1338 and @evelikov

As an aside: there's no point in pinging me. I receive all bug mail for libepoxy.

kev009:
LGTM from code inspection, we didn't experience any known crashes on FreeBSD that I am aware of but just switched to libglvnd everywhere so would like to drop the direct GLX dependency.

evelikov:
From a quick look libGL -> libOpenGL fallback sounds like the right thing to do. Thank you.

One would suggest a couple of things:
 - elaborate/reference the inline "existing behaviour" comment - either add concrete details how, or simply reference the issues/broken workloads https://github.com/anholt/libepoxy/issues/240 and https://github.com/anholt/libepoxy/issues/252
 - adding an in-tree test, with trivial reproducer of said bugs - so it is even harder to accidentally break in the future

Doubt I can test it at the moment @linkmauve - do you have some time to check this on your end?

batesj:
Does the test need to be in this pull request or can it be a followup? The nature of this change is at the level of attempting to open libGL.so.1 and libOpenGL.so.0 so a test could have have dummy libraries that confirm libGL.so.1 is loaded if it exists, and then fall back on libOpenGL.so.0.

Ernest1338:
I'm on vacation at the moment, I will test this when I get home.

evelikov:
> Does the test need to be in this pull request or can it be a followup? The nature of this change is at the level of attempting to open libGL.so.1 and libOpenGL.so.0 so a test could have have dummy libraries that confirm libGL.so.1 is loaded if it exists, and then fall back on libOpenGL.so.0.

I don't have a preference if tests come here or as follow-up.

The test you outlined doesn't sound super useful. Tests which do functional testing, similar to the bugs #240 #252 and even #253 would be much more valuable. Mind you I did not trace each use-case, so it might be that they all follow the exact same problematic pattern.

Note: the tests are a suggestion, ultimately it's not my call to make.

batesj:
Perhaps there's an easy github workflow that would repro some of the original integration bugs in #240, #252 and/or #253. There is some investigation to be done for that, so would prefer to not block this CL -- filed #258 for that.

In the mean time this CL fixes the bug that @ya-isakov described in https://github.com/anholt/libepoxy/pull/229#issue-458494975. To reiterate the motivation here: without this fix, it is impossible for a wayland-only system to use OpenGL via libepoxy.

kev009:
From the outside it seems like this is an acceptable workaround to an issue that is not very well understood.  With my cursory inspection it seems to require non-trivial knowledge of how these handles are used everywhere (which might affect library-consumers?) and whether there are known or unknown limitations with libOpenGL in a GLX setup.  Until the issue is better understood, it is difficult to prescribe a test plan, and the salient change is to preserve old known good behavior for GLX.

batesj:
> * elaborate/reference the inline "existing behaviour" comment - either add concrete details how, or simply reference the issues/broken workloads [Draw calls missing in apitrace (regression) #240](https://github.com/anholt/libepoxy/issues/240) and [Upgrading libepoxy from 1.5.5 to 1.5.7 results in Xorg crashing on boot #252](https://github.com/anholt/libepoxy/issues/252)

Done.

src/dispatch_common.c
```diff
@@ -670,13 +670,23 @@ epoxy_load_gl(void)
     get_dlopen_handle(&api.gl_handle, OPENGL_LIB, true, true);
 #else
 
+    // Prefer GLX_LIB over OPENGL_LIB to maintain existing behavior.
+    // Without this ordering, dependent projects experienced bugs such as:
```
evelikov, line 674: Tiniest of nits:

```suggestion
    // Using the inverse ordering OPENGL_LIB -> GLX_LIB, causes issues such as:
```

With that the MR is:
```
Reviewed-by: Emil Velikov <emil.velikov@collabora.com>
```

evelikov:
@batesj I'm aware how annoying libGL.so can be. In case it's unclear - I'm not suggesting we should block this MR. I did said "suggestion" and not "requirement" exactly for that reason ;-)

batesj:
Thanks for the review, @evelikov!

Ernest1338:
So I tested this, and found no issues (related to Xorg at least). ;)

batesj:
Thanks for testing @Ernest1338, sounds like we can land this then.

ebassi:
I agree. Thanks to everyone who tested this, it's very much appreciated.

Thanks @batesj for working on this, as well.

batesj:
Thanks @ebassi, let me know when you think we can release version 1.5.9 with this change.

PR #259, opened 2021-08-30
meson.build: Fallback from libGL to libOpenGL when GLX is disabled

lanodan:
This change allows to build & use libepoxy in a wayland environment without libX11 (as libGL.so depends on it for glx).

There is also a separated "glx" library but that would require more changes to the meson code and I'm not sure if it would be useful to libepoxy.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -164,6 +164,9 @@ endif
 # Dependencies
 dl_dep = cc.find_library('dl', required: false)
 gl_dep = dependency('gl', required: false)
+if not gl_dep.found() and not build_glx
+	gl_dep = dependency('opengl', required: false)
+endif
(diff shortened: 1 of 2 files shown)


danger89 (approved):

mattst88:
@ebassi:  could you merge this, please?

meson.build
```diff
@@ -164,6 +164,9 @@ endif
 # Dependencies
 dl_dep = cc.find_library('dl', required: false)
 gl_dep = dependency('gl', required: false)
+if not gl_dep.found() and not build_glx
+	gl_dep = dependency('opengl', required: false)
```
ebassi, line 168:
```suggestion
  gl_dep = dependency('opengl', required: false)
```

PR #261, opened 2021-10-06
Fix build if EGL/X11 headers are in a custom prefix

arichardson:
Currently the build works on most systems since these headers are
usually installed in the same directory as other dependencies or in
the default include directory. However, on some platforms the X11/GL
headers are installed to a different prefix, and then the build fails
due to missing includes.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -165,9 +165,11 @@ endif
 dl_dep = cc.find_library('dl', required: false)
 gl_dep = dependency('gl', required: false)
 egl_dep = dependency('egl', required: false)
+elg_headers_dep = egl_dep.partial_dependency(compile_args: true, includes: true)
 
 # Optional dependencies for tests
 x11_dep = dependency('x11', required: false)
+x11_headers_dep = x11_dep.partial_dependency(compile_args: true, includes: true)
 
 # GLES v2 and v1 may have pkg-config files, courtesy of downstream
 # packagers; let's check those first, and fall back to find_library()
--- a/src/meson.build
+++ b/src/meson.build
@@ -59,6 +59,12 @@ epoxy_deps = [ dl_dep, ]
 if host_system == 'windows'
   epoxy_deps += [ opengl32_dep, gdi32_dep ]
 endif
+if enable_x11
+  epoxy_deps += [ x11_headers_dep, ]
+endif
+if build_egl
+  epoxy_deps += [ elg_headers_dep, ]
+endif
 
 libepoxy = library(
   'epoxy',


npdgm:
Adding information for search engines.

It affects cross compilation when `sys_root` and `pkg_config_libdir` are used :
`include/epoxy/egl_generated.h:11:10: fatal error: EGL/eglplatform.h: No such file or directory`

I'm in a situation where applying patches is not convenient, and worked around it with: `-Dc_args=-I/my-sysroot/opt/stuff/include`
The whole command looks like this:
`meson build --cross-file=/build/cross-meson.txt --prefix=/opt/stuff --libdir=lib --wrap-mode=nofallback --buildtype=release -Dc_args=-I/my-sysroot/opt/stuff/include -Ddocs=false -Dglx=no -Degl=yes -Dx11=false -Dtests=false`

arichardson:
Rebased and fixed formatting. This pull request is a superset of #256, so that one can also be closed if this one is merged.

arichardson:
ping?

arichardson:
@ebassi Any chance this build fix could be merged? I'd very much like to stop maintaining a separate fork just for this compilation fix.

ebassi:
Thanks for the merge request. Could you please specify *which* platforms this use case applies to? Saying "some platforms" does help in case we need to revert in the future, when those platforms get fixed/changed.

npdgm:
Regardless of platforms, their fix is needed for working on a meson cross-build, and have EGL be found in the target rootfs. Convenient for building embedded

ebassi:
> their fix is needed for working on a meson cross-build

Then the commit message should say so.

To be fair, this is the first time I've seen using a `partial_dep()` in Meson to deal with cross-compilation, so I'm not entirely sure if the fix should reside in libepoxy or if it's just a case of those cross-compilation build environments being broken. I don't really enjoy adding workarounds inside projects for weirdness that should be confined into specific builds.

arichardson:
It affects me when cross compiling for FreeBSD but it would also apply to others.

However this is not related to cross-compilation it also affects any platform that e.g. installs the headers to say /opt/X11

 See also #256 which says it affects Broadcom platforms

arichardson:
Update the commit message with an example that shows this is needed even for non-cross-compiled environments. It just happens to work by chance right now since X11 headers are usually in `/usr/include` which is part of the default include path, but if you install them to say `/opt/X11` they will not be found.

arichardson:
@ebassi sorry to bother you again, but it would be nice if this could be merged. The only this this PR changes is that it adds `-I<X11_PREFIX>/include` and `-I<EGL_PREFIX>/include` to the compiler flags. If those prefixes are the same as ones that have already been added (e.g. `/usr/include` or `/usr/local/include`) meson will not bother adding them so it changes nothing for existing users.

ebassi:
As I said, the commit message could be better, because it took me a while to understand **why** you're using a partial dependency object—and I wrote most of the build system for libepoxy.

The issue is that you need to create a partial dependency so we don't mistakenly link libepoxy with libraries that we're going to `dlopen()` during normal operations. The commit message should make this explicit, so I don't open `meson.build` in 12 months and have to spend time reverse engineering the changes.

It would also help to not refer to "some platforms", but to be explicit about the conditions upon which these changes rest; it's not an issue inherent to "some platform", after all, but an explicit choice by the people using *any* platform.

A possible commit message could be this one:

```
Add the right include paths for EGL and X11 headers

Libepoxy currently depends on all headers living under the same prefix. This
is not necessarily true: X11 headers can live in a separate prefix, for instance
under /opt/X11. This is also the case when cross-compiling to a platform that
sets up the build environment in non-standard ways.

We could add `x11_dep` and `egl_dep` to the libepoxy target dependencies,
but that could potentially add spurious linker flags and cause libepoxy to
depend on libraries it will dlopen() during normal operations.

To avoid that case, we use a partial_dep() object from Meson, and we limit
the dependency to compiler flags and inclusion paths.
```

arichardson:
Thank you, I've used your suggested commit message.

ebassi:
Thanks for the changes!

Issue #199, opened 2019-01-18
Windows: wglMakeCurrent in Thread causes segfault


Ninja-Koala:
Making a context current in a thread causes a segfault, minimal reproducing example [here](https://gist.github.com/Ninja-Koala/33bf18903cf2ea0d3a7e2c8a1021d30b)

ebassi:
Thanks for your patience.

I don't have a Windows machine to test this on; could you please attach a stack trace for the segfault?

Ninja-Koala:
I will attach a stacktrace, but i will probably not have time for this before weekend.

Ninja-Koala:
So, here is the stacktrace:
[epoxy-stacktrace.txt](https://github.com/anholt/libepoxy/files/4080586/epoxy-stacktrace.txt)
I had to do some program counter magic to get a meaningful backtrace (see https://stackoverflow.com/a/9810389)

PR #265, opened 2021-11-16
Fix Win32 multithread dispatching bugs.

gjz010:
Using `__thread` (and `__declspec(thread)` on MSVC) to replace dirty DllMain hack and thus allowing static build (`-Ddefault_library=static`). (Possibly solving #200 issue.)
Also fixed race condition in dispatch table logic. Now on Win32 it will always use dispatch table. (#199 minimal reproducing example passed.)

### DIFF
--- a/src/dispatch_wgl.c
+++ b/src/dispatch_wgl.c
@@ -89,6 +89,8 @@ epoxy_handle_external_wglMakeCurrent(void)
     }
 }
 
+#ifndef EPOXY_STATIC_BUILD
+
 /**
  * This global symbol is apparently looked up by Windows when loading
  * a DLL, but it doesn't declare the prototype.
@@ -140,6 +142,7 @@ DllMain(HINSTANCE dll, DWORD reason, LPVOID reserved)
 
     return TRUE;
 }
+#endif
 
 WRAPPER_VISIBILITY (BOOL)
 WRAPPER(epoxy_wglMakeCurrent)(HDC hdc, HGLRC hglrc)
--- a/src/dispatch_wgl.c
+++ b/src/dispatch_wgl.c
@@ -27,8 +27,8 @@
 
 #include "dispatch_common.h"
 
-static bool first_context_current = false;
-static bool already_switched_to_dispatch_table = false;
+//static bool first_context_current = false;
+// static bool already_switched_to_dispatch_table = false;
 
 /**
  * If we can determine the WGL extension support from the current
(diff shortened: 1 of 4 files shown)


src/dispatch_wgl.c
```diff
@@ -27,8 +27,8 @@
 
 #include "dispatch_common.h"
 
-static bool first_context_current = false;
-static bool already_switched_to_dispatch_table = false;
+//static bool first_context_current = false;
+// static bool already_switched_to_dispatch_table = false;
```
ebassi, line 31: Please, don't leave commented out lines.

src/gen_dispatch.py
```diff
@@ -816,14 +816,22 @@ def write_source(self, f):
...
 
+        self.outln('#undef EPOXY_DISPATCH_PTR')
+        self.outln('#define EPOXY_DISPATCH_PTR(name) name##_dispatch_table_thunk')
+
         self.outln('static struct dispatch_table resolver_table = {')
         for func in self.sorted_functions:
             self.outln('    epoxy_{0}_dispatch_table_rewrite_ptr, /* {0} */'.format(func.wrapped_name))
         self.outln('};')
         self.outln('')
+        """
```
ebassi, line 831: You've added `EPOXY_STATIC_BUILD`, but now you've commented out the whole block using it.

src/gen_dispatch.py
```diff
@@ -843,14 +864,17 @@ def write_source(self, f):
         self.outln('}')
         self.outln('')
 
+        """
```
ebassi, line 867: Please, don't comment out blocks: either code is used, or it's removed. We have sources under revision control for a reason.

ebassi (changes requested):
Thanks for looking into this, it's very much appreciated.

src/dispatch_wgl.c
gjz010: Sorry for the dirty code. I'll try to clean them up.

gjz010:
FYI: since the patch forced using dispatch table on Win32 the compiler will complain about unused `epoxy_*_global_rewrite_ptr` functions, like:
```
include/epoxy/wgl_generated.h:956:26: warning: 'epoxy_wglWaitForSbcOML_global_rewrite_ptr' defined but not used [-Wunused-function]
  956 | #define wglWaitForSbcOML epoxy_wglWaitForSbcOML
      |                          ^~~~~~~~~~~~~~~~~~~~~~
../src/dispatch_common.h:105:5: note: in definition of macro 'GEN_GLOBAL_REWRITE_PTR_RET'
  105 |     name##_global_rewrite_ptr args                               \
      |     ^~~~
src/wgl_generated_dispatch.c:4686:1: note: in expansion of macro 'GEN_THUNKS_RET'
 4686 | GEN_THUNKS_RET(BOOL, wglWaitForSbcOML, (HDC hdc, INT64 target_sbc, INT64 * ust, INT64 * msc, INT64 * sbc), (hdc, target_sbc, ust, msc, sbc))
      | ^~~~~~~~~~~~~~
src/wgl_generated_dispatch.c:4686:22: note: in expansion of macro 'wglWaitForSbcOML'
 4686 | GEN_THUNKS_RET(BOOL, wglWaitForSbcOML, (HDC hdc, INT64 target_sbc, INT64 * ust, INT64 * msc, INT64 * sbc), (hdc, target_sbc, ust, msc, sbc))
      |                      ^~~~~~~~~~~~~~~~
```
(See CI log in https://github.com/anholt/libepoxy/runs/4226313433?check_suite_focus=true).
I would appreciate it if you could tell how this should be handled.

src/gen_dispatch.py
gjz010: The new flag is now removed since `__thread` also applies to shared library.

src/dispatch_wgl.c
gjz010: Removed.

src/gen_dispatch.py
gjz010: Removed. Please tell me if there is more improper code.

danger89:
@ebassi  Could you support him further with his question, so this can be merged eventually?

dpogue:
This pull request resolved some issues I was seeing around the dispatch returning null for function pointers when accessed from different threads

sldevel:
Is there any progress on this matter ?

I am too seeing crashes with Windows builds (no issue on Linux) when trying to create and switch to shared GL contexts, with s Second Life third party viewer...

ebassi (approved):

ebassi:
> @ebassi Could you support him further with his question, so this can be merged eventually?

One way to avoid compiler warnings would be to mark everything with `__attribute__((unused))`, but that's not really a good plan. I don't really understand why the compiler would complain about unused symbols in a library, though.

ebassi:
Since I'm not a Windows expert or developer, I'd like to get a second pair of eyes; switching to thread-local storage using compiler annotations instead of run time API is a bit iffy; I don't want to regress the shared library use case on Windows just to fix the static build case. I honestly think that static builds of C library *in 2022* are a mistake more often than not, anyway. I care about static builds insofar as they don't require weird contortions.

sldevel:
As a follow up to my previous comment, I must point out that while shared GL context do not crash any more libepoxy with that patch, I am faced with a weird issue, where the viewer process never exits on WinMain() return !
I worked around it by calling "TerminateProcess(GetCurrentProcess(), exit_code);" just before "return exit_code;" at the very end of WinMain(), but it is of course an ugly, dirty hack.
This issue also arises, when I configure the viewer not to create shared GL contexts beside the main one (but of course still flags the main context as shared and still calls wglCreateContext() for the main context).

PR #266, opened 2021-12-06
Fix static build (DllMain)

danger89:
Do *not* include the `DllMain` entry point in the static build, only when it's a shared build.

### DIFF
--- a/src/dispatch_wgl.c
+++ b/src/dispatch_wgl.c
@@ -89,6 +89,7 @@ epoxy_handle_external_wglMakeCurrent(void)
     }
 }
 
+#ifdef DLL_EXPORT
 /**
  * This global symbol is apparently looked up by Windows when loading
  * a DLL, but it doesn't declare the prototype.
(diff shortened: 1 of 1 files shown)


danger89:
@ebassi Maybe you want to look at this PR?

ebassi:
I receive mail for all PRs/issues; I don't always have time to review things immediately: libepoxy is not the only project I work on.

ebassi:
It would be nice to have some test, or a CI pipeline job, that checks the static build on Windows and ensures that a) it works and b) it does not regress.

danger89:
> I receive mail for all PRs/issues; I don't always have time to review things immediately: libepoxy is not the only project I work on.

I understand, no blame.

> It would be nice to have some test, or a CI pipeline job, that checks the static build on Windows and ensures that a) it works and b) it does not regress.

I agree. I guess, we can extend the github action workflow file with a static build: 
https://github.com/anholt/libepoxy/blob/master/.github/workflows/msys2.yml

danger89:
Ping? (sorry to just bother you, otherwise I don't see this getting merged)

danger89:
Any feedback?

hesham-essam:
I would appreciate if this is supported upstream as well.

dg0yt:
Is it really enough to not compile DllMain? Wouldn't it take an alternative way to do the initialization which is in the body of this function?

PR #270, opened 2022-04-07
test: Fix dlwrap on riscv64

piggynl:
See [sysdeps/unix/sysv/linux/riscv/rv64/libdl.abilist](https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/riscv/rv64/libdl.abilist;h=33ff573df8748197102314b6ca7579e060c7b8ad) in glibc's source.

### DIFF
--- a/test/dlwrap.c
+++ b/test/dlwrap.c
@@ -237,6 +237,7 @@ dlwrap_real_dlsym(void *handle, const char *name)
          * In the meantime, I'll just keep augmenting this
          * hard-coded version list as people report bugs. */
         const char *version[] = {
+            "GLIBC_2.27",
             "GLIBC_2.17",
             "GLIBC_2.4",
             "GLIBC_2.3",


andreabolognani:
Hi, can the maintainers please take a look at this patch?

I can't claim that I fully understand what it does, but I can confirm that applying it on top of 1.5.10 makes it possible to successfully build the Fedora package on riscv64.

The approach currently used on Fedora (where riscv64 is not yet an officially supported architecture) and Debian (where it already is) is to simply skip the test suite altogether, which is clearly not ideal. See the [Fedora hack](http://fedora.riscv.rocks:3000/rpms/libepoxy/src/commit/3eff20ef55a107d7f411f7cbc9945b9f47a68bdb/libepoxy.spec#L45-L55) and the [Debian one](https://salsa.debian.org/xorg-team/lib/libepoxy/-/blob/40240898cfffd32911a23908eb91842f31ef2ac5/debian/rules#L10-16) for reference.

There is a [PR](https://src.fedoraproject.org/rpms/libepoxy/pull-request/2) that is trying to get the patch added to the Fedora package: this is much more likely to move forward if the patch in question had already been accepted upstream. And of course we want a future upstream release to work out of the box.

Thanks in advance for your consideration!

andreabolognani:
@ebassi any chance you could look into this? It would be really great to have this fix (or an equivalent one) merged so that we can build and test libepoxy properly on riscv64. Thanks!

andreabolognani:
@ebassi another friendly ping :)

ebassi:
I don't have access to any RISC-V board, so I can't validate the pull request; all I can do is trust the fix, and merge it, which I plan to do for the next release of libepoxy. I have been busy with `${DAY_JOB}`, and there have been a few other pull requests on the queue.

andreabolognani:
@ebassi no worries, we're all busy one way or another so the intent behind my pings was simply to ensure that this wouldn't fall between the cracks :)

If you don't have any riscv64 hardware handy, you can go the emulation route. It's going to require some patience since things are obviously going to be a bit slow, but it should be relatively straightforward. The process is documented [on the Fedora wiki](https://fedoraproject.org/wiki/Architectures/RISC-V/QEMU) and I'm happy to provide additional assistance should you need it.

andreabolognani:
It's been a couple months now. Any progress?

PR #298, opened 2024-09-10
format: trim spaces and use space instead tab

lygstate:


ebassi:
Thanks, but I'd rather not have commits that do this independently; I'd prefer that fixes to specific bits ended up also fixing the whitespace. This makes it easier to deal with `git blame` and logs.

Issue #290, opened 2023-07-30
CMake configuration file


yanxiang-wang:
Hello, when I install this library using Vcpkg. They cannot show the note on how to use this library.
The reason is that the upstream didn't provide the Cmake configuration file 

[https://github.com/microsoft/vcpkg/issues/32651](https://github.com/microsoft/vcpkg/issues/32651)

ebassi:
This is not a libepoxy issue: we're not going to add a CMake build to the project.

The vcpkg project should use the supported Meson build; failing that, they can maintain a CMake build downstream, though of course that also means they will need to support it themselves.

Since your issue is finding the installed libepoxy, you should use `FindPkgConfig(epoxy)`, as libepoxy relies on the pkg-config project to define compiler and linker flags.

PR #303, opened 2025-03-14
CMake build system

Treata11:
Added CMake build system as a more powerful alternative for Meson. 
Not to mention many of the gains CMake brings; I personally needed it to build for iOS-derived platforms & thought that I contribute it to the source as many people would prefer CMake over any other build systems... 
Projects depending on Epoxy could also benefit from this addition.

I've mirrored the original Meson flags in this (some are incomplete such as the tests & docs). But the main part is done & I successfully built and tested it for both macOS & iOS targets & I'm pretty certain that it works for Linux as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 aclocal.m4
 autom4te.cache/
 autoscan.log
+build*/
 ChangeLog
 compile
 config.guess
@@ -56,6 +57,7 @@ src/sna/brw/brw_test
 core
 *.dll
 *.exe
+*.framework
 *-ISO*.bdf
 *-JIS*.bdf
 *-KOI8*.bdf
@@ -71,6 +73,7 @@ core
 *.pdb
 *.tar.bz2
 *.tar.gz
+*.xcframework
 #
 #		Add & Override patterns for gldispatch
 #
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,12 @@ cmake_minimum_required(VERSION 3.18)
 project(libepoxy
     VERSION 1.5.11
     LANGUAGES C
+    DESCRIPTION "GL dispatch library"
 )
+
+# Version components
+set(EPOXY_SHORT_VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR})
+set(EPOXY_FULL_VERSION ${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}.${PROJECT_VERSION_PATCH})
 
 # Set C standard and default build type
 set(CMAKE_C_STANDARD 99)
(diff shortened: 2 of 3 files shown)


Philanatidae:
I've only looked at this briefly, but one thing I noticed is you have the defines for `ENABLE_GLX`, `ENABLE_EGL`, and `ENABLE_X11` always as `1` instead of programmatically having these be defined or not based on whether or not they're available (config.h.in):
```
* Platform-specific features */
#cmakedefine ENABLE_GLX 1
#cmakedefine ENABLE_EGL 1
#cmakedefine ENABLE_X11 1
```

With that said, the maintainers of libepoxy haven't expressed interest in adding (and therefore maintaining) a second build system (see Issue #290), which is totally valid. CMake, while common, isn't a standard, and Meson is a great build tool (just doesn't work for me).

I also use CMake for my projects, so I have a fork of libepoxy with a CMakeLists.txt that I use instead. See [here](https://github.com/Philanatidae/libepoxy). Currently it works with macOS, Android, and emscripten, but feel free to PR it to add missing features if you want.

Treata11:
> I've only looked at this briefly, but one thing I noticed is you have the defines for ENABLE_GLX, ENABLE_EGL, and ENABLE_X11 always as 1

Yes, it's pretty much incomplete ... 


> With that said, the maintainers of libepoxy haven't expressed interest in adding (and therefore maintaining) a second build system (see Issue https://github.com/anholt/libepoxy/issues/290), which is totally valid. CMake, while common, isn't a standard, and Meson is a great build tool (just doesn't work for me).

Adding an additional build system as common as CMake is actually a substantial move. Bet there were twice as stars if this was a CMake configured repo. And, maintaining it wouldn't be any burden at all. Any changes to Meson could be directly mirrored to CMake, pretty easily in no time & even if the maintainers are unable to that, many contributors would bump into the issue & fix it in a PR... That's how things are suppose to work around here AFAIK :) 
By the way, there's no such thing as **"Standard"**  ... They're all conventions & practices.


> I also use CMake for my projects, so I have a fork of libepoxy with a CMakeLists.txt that I use instead. See [here](https://github.com/Philanatidae/libepoxy). Currently it works with macOS, Android, and emscripten, but feel free to PR it to add missing features if you want.

Didn't the maintainers approve your PR? 
I've already fixed my issue of cross-compilation. Would love to add the changes to your fork since yours seems to be more complete. 
Guess at least an open/unmerged PR would help with the visibility if the owners insist to say no to CMake & you've tried before :)

ebassi:
No, thank you. I am not going to maintain a build system I don't use, and especially one I **really** do not like. Issue #290 already explains the reason.

> Bet there were twice as stars if this was a CMake configured repo.

Open source software is not maintained via entirely useless popular vote metrics like "github stars".

> And, maintaining it wouldn't be any burden at all.

This tells me you have no practical experience with maintaining projects that sit very low in the stack.

Treata11:
And it's clear to me that you're not even getting the point here :)

Well, I'm not going to enter a unfruitful argument with a stubborn person to kill time. 
Your repo, your right to decide for it. I got what I came for & thought it might help someone else.

PR #311, opened 2025-10-25
Fix EGL support on macOS

mohd-akram:
This is a simpler version of #239 and is the patch we're using in MacPorts.

### DIFF
--- a/src/dispatch_common.h
+++ b/src/dispatch_common.h
@@ -28,7 +28,7 @@
 #define PLATFORM_HAS_GLX ENABLE_GLX
 #define PLATFORM_HAS_WGL 1
 #elif defined(__APPLE__)
-#define PLATFORM_HAS_EGL 0 
+#define PLATFORM_HAS_EGL ENABLE_EGL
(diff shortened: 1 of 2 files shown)


osy:
@ebassi Any chance this one can be reviewed? EGL support for macOS in QEMU is blocked on libepoxy support

ebassi:
The commit message is very underwhelming.

Why does removing the symbols from `dispatch_common.h` and changing the shared libraries in `dispatch_common.c` "fixes" macOS?

src/dispatch_common.h
```diff
@@ -27,10 +27,6 @@
 #define PLATFORM_HAS_EGL ENABLE_EGL
 #define PLATFORM_HAS_GLX ENABLE_GLX
 #define PLATFORM_HAS_WGL 1
-#elif defined(__APPLE__)
```
ebassi: I'd rather keep this section, but change the values, instead of removing it and relying on the defaults. The defaults are generally understood to apply to Linux.

